PR #8, opened 2023-09-11
feat: compatibility with pydantic v2.x.y

ddneilson:

Pydantic version 2 was released, and contains pydantic v1 as a submodule. This change allows this library to function with both major versions of Pydantic; it will automatically select the v1 submodule if v2 is available in the environment.

Try to import the v1 module from pydantic v2, and fallback to pydantic v1 if that fails.

Downstream consumers will be open to use Pydantic V2 at any time without needing to wait for this library to be updated.

Using a `hatch shell`, I ran the test suite with both pydantic v2.3.0 and then again with v1.10.11 installed (via pip install --force).

N/A

No, quite the opposite.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ requires-python = ">=3.9"
 
 dependencies = [
   "pyyaml ~= 6.0",
-  "pydantic ~= 1.10",
+  "pydantic >= 1.10",
 ]
 
 [tool.hatch.build]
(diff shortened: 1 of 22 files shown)


ddneilson:
Closing in favour of an upcoming PR. Pydantic 1.10.17+ contain the `pydantic.v1` namespace to make v2 compatibility much easier to do.

PR #96, opened 2024-03-27
feat: add validation that associative op's args are equal length

ddneilson:

The Open Job Description specification says "Every comma-separated expression within an associative operator must have the exact same number of values defined in their range." The current model validators to not ensure that this is the case.

Reference: https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#343-combinationexpr

I've implemented `validate_step_parameter_space_dimensions()` to validate this constraint and added a validator to the StepParameterSpace to use it. My first inclination was to implement this validation as part of constructing the StepParameterSpaceIterator since it already has a recursive expression parse, but that is a chicken & egg problem; we need a StepParameterSpace to construct a StepParameterSpaceIterator, but we don't have one yet when doing validation for the construction of a StepParameterSpace.

The model validator enforces a previously neglected constraint.

I have added unit tests that fully cover the new functionality.

Yes, in the official specification

No

----

### DIFF
--- a/src/openjd/model/v2023_09/_model.py
+++ b/src/openjd/model/v2023_09/_model.py
@@ -32,6 +32,7 @@ from .._capabilities import (
 )
 from .._internal import (
     CombinationExpressionParser,
+    validate_step_parameter_space_dimensions,
     validate_unique_elements,
 )
 from .._internal._variable_reference_validation import (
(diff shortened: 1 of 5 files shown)


src/openjd/model/v2023_09/_model.py
```diff
@@ -729,6 +730,25 @@ class StepParameterSpace(OpenJDModel_v2023_09):
     taskParameterDefinitions: dict[Identifier, TaskRangeParameter]
     combination: Optional[CombinationExpr] = None
 
+    @validator("combination")
+    def _validate_parameter_space(cls, v: str, values: dict[str, Any]) -> str:
+        if v is None:
+            return v
```
jericht, line 736:
```suggestion
    def _validate_parameter_space(cls, v: Optional[str], values: dict[str, Any]) -> str:
        if v is None:
            return v
```

nit: Can `v` be `None`? Type hint should be updated to reflect it

jericht (approved):

jusiskin (approved):

PR #138, opened 2024-08-29
docs: enhance contributing guidelines to be more accessible

ddneilson:

The contributing and development guidelines in this repository could benefit from an improvement pass to make workflows, expectations, and intentions clearer.

Rework/refactor the CONTRIBUTING.md, DEVELOPMENT.md, GitHub issue templates, and pull request template with the goal of improving understandability to those not familiar with the code base, hatch-based workflows, GitHub pull request workflows, and our use of conventional commits

The hope is that the development and contribution process should be much clearer for everyone.

N/A

It is documentation, so yes.

No.

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,62 +6,106 @@ documentation, we greatly value feedback and contributions from our community.
 Please read through this document before submitting any issues or pull requests to ensure we have all the necessary
 information to effectively respond to your bug report or contribution.
 
+Table of contents:
+
+* [Reporting Bugs/Feature Requests](#reporting-bugsfeature-requests)
+* [Development](#development)
+    * [Finding contributions to work on](#finding-contributions-to-work-on)
+    * [Talk with us first](#talk-with-us-first)
+    * [Contributing via Pull Requests](#contributing-via-pull-requests)
+    * [Conventional Commits](#conventional-commits)
+* [Licensing](#licensing)
+
 ## Reporting Bugs/Feature Requests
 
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -6,62 +6,106 @@ documentation, we greatly value feedback and contributions from our community.
...
+
+* [Reporting Bugs/Feature Requests](#reporting-bugsfeature-requests)
+* [Development](#development)
+    * [Finding contributions to work on](#finding-contributions-to-work-on)
+    * [Talk with us first](#talk-with-us-first)
+    * [Contributing via Pull Requests](#contributing-via-pull-requests)
+    * [Conventional Commits](#conventional-commits)
+* [Licensing](#licensing)
+
 ## Reporting Bugs/Feature Requests
```
crowecawcaw, line 19: Somewhere in this file it might be useful to point users to the OJD spec repository if they have an OJD request. This repo is just for the Python implementation of the spec.

crowecawcaw (approved):

CONTRIBUTING.md
ddneilson: I believe that's covered by the intro paragraph of the README. Do you think that some additional clarification should be added there?

crowecawcaw: I could see someone working with the Python library, wishing it supported something, then raising an issue here that really would be a spec update. But maybe the README covers it. We can always add something later if people do raise issues in the wrong place.

epmog (approved):

PR #164, opened 2025-02-08
chore!: Update model parsing to use Pydantic V2 instead of Pydantic V1

mwiebe:

The RFC 0002 proposal https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0002-model-extensions.md
depends on parsing differently based on the contents of the "extensions" field at the root of the document.
Pydantic V1 doesn't provide a way to carry a context to all the validators, while Pydantic V2 does.

Update job and environment parsing to use Pydantic V2.

We will be able to implement the extensions RFC. Pydantic V2 is a nice upgrade over Pydantic V1, so
further work in this library can benefit from upstream improvements.

Yes, have run all the tests. In some cases, I found that Pydantic V2 has more strict behavior than V1,
and I updated the tests to accommodate. While I mostly preserved the error message formatting, I
made some small tweaks to the messages being validated in the tests.

Yes, I updated the explainer documentation that included Pydantic V1 details to reflect V2.

While largely compatible, the underlying implementation uses Pydantic V2 models instead of Pydantic V1
models, and they have incompatible interfaces. Usage of this library that depends on those internal details
will need to be updated.

This change doesn't impact security. Updating to Pydantic V2 will help the library to maintain up-to-date
dependencies.

----

### DIFF
--- a/src/openjd/model/v2023_09/_model.py
+++ b/src/openjd/model/v2023_09/_model.py
@@ -6,23 +6,23 @@ import re
 from decimal import Decimal, InvalidOperation
 from enum import Enum
 from graphlib import CycleError, TopologicalSorter
-from typing import TYPE_CHECKING, Any, ClassVar, Literal, Optional, Type, Union, cast
-from typing_extensions import Annotated
+from typing import Any, ClassVar, Literal, Optional, Type, Union, cast
+from typing_extensions import Annotated, Self
 
-from pydantic.v1 import (
+from pydantic import (
+    field_validator,
+    model_validator,
+    StringConstraints,
     Field,
     PositiveInt,
     PositiveFloat,
     StrictBool,
     StrictInt,
     ValidationError,
-    conint,
-    conlist,
-    constr,
-    root_validator,
-    validator,
+    ValidationInfo,
 )
-from pydantic.v1.error_wrappers import ErrorWrapper
+from pydantic_core import InitErrorDetails
+from pydantic.fields import ModelPrivateAttr
 
 from .._format_strings import FormatString
 from .._errors import ExpressionError, TokenError
@@ -119,10 +119,10 @@ class ValueReferenceConstants(Enum):
 #    C1 = 0x80-0x9F
 #         https://www.unicode.org/charts/PDF/U0080.pdf
 _Cc_characters = r"\u0000-\u001F\u007F-\u009F"
-_standard_string_regex = rf"(?-m:^[^{_Cc_characters}]+\Z)"
+_standard_string_regex = rf"(?-m:^[^{_Cc_characters}]+\z)"
 
 # Latin alphanumeric, starting with a letter
-_identifier_regex = r"(?-m:^[A-Za-z_][A-Za-z0-9_]*\Z)"
+_identifier_regex = r"(?-m:^[A-Za-z_][A-Za-z0-9_]*\z)"
 
 # Regex for defining file filter patterns allowed for use in file dialogs.
 # 1. Allowable values: "*", "*.*", and "*.[:file-extension-chars:]+".
@@ -136,7 +136,7 @@ _file_dialog_filter_pattern_regex = (
     rf"(?-m:^(?:\*|\*\.\*|\*\."
     rf"[^{_Cc_characters}\\/\*"
     rf"\?\[\]#%&\{{\}}<>\$\!'"
-    rf"\\\":@`|=]+)\Z)"
+    rf"\\\":@`|=]+)\z)"
 )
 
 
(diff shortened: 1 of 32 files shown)


Ahuge:
🎉

crowecawcaw (dismissed):

src/openjd/model/v2023_09/_model.py
```diff
@@ -119,10 +119,10 @@ class ValueReferenceConstants(Enum):
 #    C1 = 0x80-0x9F
 #         https://www.unicode.org/charts/PDF/U0080.pdf
 _Cc_characters = r"\u0000-\u001F\u007F-\u009F"
-_standard_string_regex = rf"(?-m:^[^{_Cc_characters}]+\Z)"
+_standard_string_regex = rf"(?-m:^[^{_Cc_characters}]+\z)"
```
epmog, line 122: Huh, this surprised me. I know there are lower-case AND upper-case escaped characters in the python regex syntax that differ in functionality, but both `\z` and `\Z` match absolute end of string? Documentation only shows `\Z`: https://docs.python.org/3/library/re.html#regular-expression-syntax

src/openjd/model/v2023_09/_model.py
```diff
@@ -177,24 +183,21 @@ class JobTemplateName(FormatString):
 class CommandString(FormatString):
     _min_length = 1
     # All unicode except the [Cc] (control characters) category
-    _regex = f"(?-m:^[^{_Cc_characters}]+\Z)"
+    _regex = f"(?-m:^[^{_Cc_characters}]+\\Z)"
```
epmog, line 186: Previously a bug? Indicating a missing test? Same with ArgString

src/openjd/model/v2023_09/_model.py
```diff
@@ -619,29 +644,57 @@ class FloatTaskParameterDefinition(OpenJDModel_v2023_09):
...
-        return v
+        cls, value: list[Union[Decimal, TaskParameterStringValue]]
+    ) -> list[Union[Decimal, TaskParameterStringValue]]:
+        errors = list[InitErrorDetails]()
+        for i, item in enumerate(value):
+            if isinstance(item, TaskParameterStringValue):
+                # A TaskParameterStringValue is a FormatString.
+                # FormatString.expressions is the list of all expressions in the format string
+                # ( e.g. "{{ Param.Foo }}").
+                # Reject the string if it contains any expressions.
```
epmog: nit: The behaviour hasn't changed and the comment is the same as it was previously, but I believe this comment is wrong (and other similar places)?

We aren't rejecting the string if it contains any expressions, we just return the value as it was.

test/openjd/model/test_step_param_space_iter.py
```diff
@@ -32,7 +32,7 @@ class TestStepParameterSpaceIterator_2023_09:  # noqa: N801
     @pytest.mark.parametrize(
         "range_int_param",
         [
-            RangeListTaskParameterDefinition_2023_09(type=ParameterValueType.INT, range=[1, 2]),
+            RangeListTaskParameterDefinition_2023_09(type=ParameterValueType.INT, range=["1", "2"]),
```
epmog, line 35: Is this closer to how they're actually used internally or a stylistic choice? Do we still have some tests when they're not strings/should we have some?

epmog:
And the follow-up, it doesn't seem like it's breaking, but because it's such an overhaul underneath we're going to go with a minor version bump? I think I can get behind that.

src/openjd/model/v2023_09/_model.py
mwiebe: It produces a warning instead of an error, but still worth fixing:

```
>>> "\Z"
<stdin>:1: SyntaxWarning: invalid escape sequence '\Z'
'\\Z'
>>> "\\Z"
'\\Z'
```

src/openjd/model/v2023_09/_model.py
mwiebe: This is because of https://docs.pydantic.dev/latest/migration/#patterns-regex-on-strings - the validations being done by the Pydantic rust code use `\z`, while the validations being done in Python code stayed as `\Z`.

test/openjd/model/test_step_param_space_iter.py
mwiebe: In the definition of the type being constructed, range is a list of strings. I don't see this difference described in the migration guide, but found that in V1 it allows integer values even though the type was str, but in V2 it doesn't.

mwiebe:
> And the follow-up, it doesn't seem like it's breaking, but because it's such an overhaul underneath we're going to go with a minor version bump? I think I can get behind that.

You mean in https://github.com/OpenJobDescription/openjd-cli/pull/134, right? It seemed reasonable to err that way in an abundance of caution, yes.

src/openjd/model/v2023_09/_model.py
mwiebe: I updated the comment to this:

```
# Validate the string if it does not contain any expressions, in order to catch
# errors earlier when possible..
```

src/openjd/model/v2023_09/_model.py
epmog: Ahhh, right, the big change of v1 -> v2, thanks!

epmog (approved):
This is exciting - Ship it!

crowecawcaw (approved):

PR #167, opened 2025-02-11
feat: Implement the extensions RFC 0002

mwiebe:

https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0002-model-extensions.md

Implement support  into template parsing.

RFC 0001 and other later RFCs can be implemented in a manner conformant with RFC 0002. The base 2023-09 specification stays the same, except when a list of names for `extensions` is provided in the template, and the implementation processing the template supports those extensions.

Yes

Yes

It's not a breaking change. the new parameter `supported_extension` for parsing a template is optional, and if none are provided, it uses the base specification with no extensions enabled.

This does not impact the threat model, and the new extensions field and parameter follow the same level of input validation relevant for existing threats.

----


src/openjd/model/v2023_09/_model.py
```diff
@@ -52,8 +52,44 @@
...
+
+        OpenJDModelSubclass.model_validate(data, context=ModelParsingContext())
+
+    Individual validators receive this value as ValidationInfo.context.
+    """
+
+    extensions: set[str]
+    """Initially, is the set of supported extension names. When the 'extensions'
+    field of the template is processed, this becomes the set of extensions that
+    the the template requested."""
```
epmog: nit
```suggestion
    extensions: set[str]
    """Initially it is the set of supported extension names. When the 'extensions'
    field of the template is processed, this becomes the set of extensions that
    the template requested."""
```
Aside: is it obvious when this context changes... it's context for users of the library? Feels a little smelly to change the meaning of the variable on the instance?

Also a little weird that this class is public but we're describing behaviour that is more about how we're interacting with the object rather than anything about the object itself

test/openjd/model/test_parse.py
```diff
@@ -185,3 +188,94 @@ def test_success(self, template: dict[str, Any], expected_class: Type[OpenJDMode
...
+                },
+            },
+            "EnvironmentTemplate",
+            decode_environment_template,
+            id="environment template",
+        ),
+    ],
+)
+def test_template_extensions_list(template, template_type, decode_function) -> None:
+    with patch.object(openjd.model.v2023_09._model, "ExtensionName", MockExtensionName):
```
epmog, line 241: can we also include a separate test case that validates only using 1 valid extension, when more than 1 valid extension exist?

ie. template uses 1, model says 2 are supported

src/openjd/model/v2023_09/_model.py
```diff
@@ -2393,6 +2432,42 @@ class JobTemplate(OpenJDModel_v2023_09):
...
+    @classmethod
+    def _permitted_extension_names(
+        cls, value: Optional[ExtensionNameList], info: ValidationInfo
+    ) -> Optional[ExtensionNameList]:
+        context = cast(ModelParsingContext, info.context)
+        if value is not None:
+            # Before processing the extensions field, context.extensions is the list of supported extensions.
+            # Take the intersection of the input supported extensions with what is implemented
+            # in this list, as the implementation needs to support an extension for it to be supported.
+            supported_extensions = context.extensions.intersection(cls.supported_extension_names())
```
epmog, line 2463: At this point are they not actually the same and this does nothing? `value` is the input list, `context.extensions` IS the supported extension names until we modify it below

epmog (changes requested):

src/openjd/model/_parse.py
```diff
@@ -203,12 +227,22 @@ def decode_template(*, template: dict[str, Any]) -> JobTemplate:
...
 
+    This function places no restriction on the version of the specification. The caller
+    can inspect the `specificationVersion` property of the returned object to validate this.
+
+    By default, no extensions are supported. The caller can opt in to specific extensions,
+    by providing them as a list.
+
     Args:
         template (dict[str, Any]): An Environment Template as a dictionary object.
+        supported_extensions (list[str]): A list of extension names to support. This list is intersected
```
leongdl: nit `list[str]` is `Optional[Iterable[str]` :)

src/openjd/model/_parse.py
```diff
@@ -141,12 +153,22 @@ def decimal_to_str(data: Union[dict[str, Any], list[Any]]) -> None:
...
 
+    This function places no restriction on the version of the specification. The caller
+    can inspect the `specificationVersion` property of the returned object to validate this.
+
+    By default, no extensions are supported. The caller can opt in to specific extensions,
+    by providing them as a list.
+
     Args:
         template (dict[str, Any]): A Job Template as a dictionary object.
+        supported_extensions (list[str]): A list of extension names to support. This list is intersected
```
leongdl: nit,`list[str]` type is `Optional[Iterable[str]`

src/openjd/model/v2023_09/_model.py
```diff
@@ -2393,6 +2432,42 @@ class JobTemplate(OpenJDModel_v2023_09):
...
+                raise ValueError(
+                    f"Unsupported extension names: {', '.join(sorted(unsupported_extensions))}"
+                )
+
+            # After processing the extensions field, context.extensions is the list of
+            # extension names used by the template.
+            context.extensions = set(value)
+        else:
+            context.extensions = set()
+        return value
```
crowecawcaw, line 2476: It looks funny that we're mutating the input parameter and returning a value. Is this a Pydantic pattern?

Also should `context.extensions` be a list instead of a set?

src/openjd/model/v2023_09/_model.py
mwiebe: It's context for the parsing to use. I'll give this comment more of an overall to explain that.

test/openjd/model/test_parse.py
mwiebe: Sure

src/openjd/model/v2023_09/_model.py
mwiebe: It's the list of supported extensions that were requested in the call of the parse_job_template function. I'll update the comment to state that.

src/openjd/model/v2023_09/_model.py
mwiebe: This is a necessary part of the context-sensitive parsing that it's doing. The parsing is driven by Pydantic, which calls the validators of each field in turn. When it validates the 'extensions' field, the validator populates the context with the extensions that were requested by the doc. Then the rest of the parsing can check which ones are enabled. It's a set so that those validators can use the efficient `in` implementation.

crowecawcaw (approved):

epmog (approved):

src/openjd/model/_parse.py
```diff
@@ -189,7 +211,9 @@ def decode_job_template(*, template: dict[str, Any]) -> JobTemplate:
         )
 
     if schema_version == TemplateSpecificationVersion.JOBTEMPLATE_v2023_09:
-        return parse_model(model=JobTemplate_2023_09, obj=template)
+        return parse_model(
+            model=JobTemplate_2023_09, obj=template, supported_extensions=supported_extensions
```
leongdl, line 215: Just a question, will we change this in the near future? 2025_03 or something like that?

leongdl (commented):
LGTM

src/openjd/model/_parse.py
mwiebe: No, I think it's better to accumulate a series of extensions before starting a new revision.

PR #168, opened 2025-02-11
feat: Implement the task chunking RFC 0001

mwiebe:

https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md

Implement the task chunking additions as the TASK_CHUNKING extensions, using the context-sensitive parsing introduced by the Pydantic V2 conversion and extensions RFC implementation.

Job templates that use the task chunking feature can be parsed successfully.

Added tests for the new functionality.

Yes

Yes

Not a breaking change, it adds support for the TASK_CHUNKING extension.

This does not impact the threat model, and the new extensions field and parameter follow the same level of input validation relevant for existing threats.

----


src/openjd/model/v2023_09/_model.py
```diff
@@ -797,11 +851,134 @@ class PathTaskParameterDefinition(OpenJDModel_v2023_09):
...
+                            ctx={
+                                "error": ValueError("Value must be an integer or integer string.")
+                            },
+                            input=item,
+                        )
+                    )
+            if errors:
+                raise ValidationError.from_exception_data(cls.__name__, line_errors=errors)
+        elif isinstance(value, RangeString):
+            # TODO: nothing to do - it's guaranteed to be a format string at this point
```
leongdl: If this isn't a todo, then its just a comment ?

leongdl: IE, we can remove TODO:

test/openjd/model/v2023_09/test_create.py
```diff
@@ -394,3 +396,119 @@ def test(self) -> None:
...
+        #  2) Every entity and every field that exists is defined at least once.
+        #  3) Testing of _internal.create_job covers corner cases & exceptions; we don't worry
+        #     about those here.
+
+        # GIVEN
+        extra_kwargs = {"$schema": "blah "}  # special snowflake due to field naming
+        template = JobTemplate(
+            **extra_kwargs,
+            specificationVersion="jobtemplate-2023-09",
+            extensions=["TASK_CHUNKING"],
```
leongdl, line 416: Can we add a templated test to test with an empty array ?

src/openjd/model/v2023_09/_model.py
mwiebe: Good point, I'll fix that and where I got it from.

test/openjd/model/v2023_09/test_create.py
mwiebe: That case is already tested here: https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/test/openjd/model/test_parse.py#L254

This file test_create.py contains larger end-to-end tests for converting an already-validated JobTemplate object into a Job. What error cases are you thinking about?

leongdl: Just thinking to cover the different range of inputs, null (before), [] (empty new), populated (this test case).

Cool, if its covered then all good!

leongdl (approved):

epmog (approved):

src/openjd/model/v2023_09/_model.py
```diff
@@ -553,6 +561,52 @@ def _validate_range_expression(cls, value: Any) -> Any:
...
+    _job_creation_metadata = JobCreationMetadata(
+        resolve_fields={"defaultTaskCount", "targetRuntimeSeconds"},
+    )
+
+    @field_validator("defaultTaskCount")
+    @classmethod
+    def _validate_default_task_count(cls, value: Any) -> Any:
+        if isinstance(value, FormatString):
+            # If the string value has no expressions, can validate the value now.
+            # Otherwise will validate when
```
jusiskin, line 585: nit: this comment is unfinished

test/openjd/model/v2023_09/test_chunk_int_task_parameter_type.py
```diff
@@ -0,0 +1,658 @@
...
+                "type": "CHUNK[INT]",
+                "range": "1-100",
+                "chunks": {
+                    "defaultTaskCount": 10,
+                    "targetRuntimeSeconds": "{{Param.TargetChunkRuntime}}",
+                    "rangeConstraint": "NONCONTIGUOUS",
+                },
+            },
+            id="targetRuntimeSeconds is str expression",
+        ),
```
jusiskin, line 159: nit: Should we also have a case where `defaultTaskCount` is a format string? I see it is already covered in `test/openjd/model/v2023_09/test_create.py`, but maybe for completeness it would be good to have here as well.

test/openjd/model/v2023_09/test_chunk_int_task_parameter_type.py
```diff
@@ -0,0 +1,658 @@
...
+                "range": [1.1],
+                "chunks": {
+                    "defaultTaskCount": 10,
+                    "targetRuntimeSeconds": 1000,
+                    "rangeConstraint": "NONCONTIGUOUS",
+                },
+            },
+            "Value must be an integer or integer string.",
+            1,
+            id="disallow floats",
```
jusiskin, line 309: nit: should this mention "range"? same applies to test cases below this

src/openjd/model/v2023_09/_model.py
```diff
@@ -797,11 +851,134 @@ class PathTaskParameterDefinition(OpenJDModel_v2023_09):
...
+                                    type="value_error",
+                                    loc=(i,),
+                                    ctx={
+                                        "error": ValueError(
+                                            "String literal must contain an integer."
+                                        )
+                                    },
+                                    input=item,
+                                )
+                            )
```
jusiskin, line 961: nit: this code comment doesn't match the implementation. It seems we are only rejecting strings if they have no expressions and the string cannot be parsed as an integer.

jusiskin (commented):
Great job. Just a few nitpicks, but I don't consider them blocking. Feel free to address them or merge as-is.

jusiskin (approved):

mwiebe:
> Great job. Just a few nitpicks, but I don't consider them blocking. Feel free to address them or merge as-is.

Thanks! I'll merge this and then open a smaller PR to fix up what you found.

PR #169, opened 2025-02-14
refactor: Refactor to improve comments and reduce code duplication

mwiebe:

Improvements following up from https://github.com/OpenJobDescription/openjd-model-for-python/pull/168,
and other cleanup/code duplication reduction.

* Comments are more clear about difference between model parsing, when values may not be available, and job instantiation, when all values must be available for format strings.
* Created a new file to hold general validator functions for some commonly used patterns.
* Adjust the model to reduce the number of confusingly spurious errors, like when errors include "must be int" and "must be string" for the exact same field, seemingly contradictory errors.

Better experience with the code base and the error messages.

Ran the unit tests.

Code comments improved.

No.

No.

----

### DIFF
--- a/src/openjd/model/_internal/_validator_functions.py
+++ b/src/openjd/model/_internal/_validator_functions.py
@@ -1,0 +1,119 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from typing import cast, Any, Callable, Union, Optional
+from decimal import Decimal, InvalidOperation
+
+from pydantic_core import PydanticKnownError, ValidationError, InitErrorDetails
+
+from .._format_strings import FormatString
+
+
+def validate_int_fmtstring_field(
+    value: Union[int, float, Decimal, str, FormatString], ge: Optional[int] = None
+) -> Union[int, float, Decimal, str, FormatString]:
+    """Validates a field that is allowed to be either an integer, a string containing an integer,
+    or a string containing expressions that resolve to an integer."""
+    value_type_wrong_msg = "Value must be an integer or a string containing an integer."
+    # Validate the type
+    if isinstance(value, str):
+        if not isinstance(value, FormatString):
+            value = FormatString(value)
+        # If the string value has no expressions, we can validate the value now.
+        if len(value.expressions) == 0:
+            try:
+                int_value = int(value)
+            except ValueError:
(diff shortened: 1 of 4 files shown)


leongdl (approved):
Awesome refactoring!

test/openjd/model/v2023_09/test_chunk_int_task_parameter_type.py
```diff
@@ -304,9 +307,39 @@ def test_chunk_int_task_parameter_parse_success(data: dict[str, Any]) -> None:
...
-            "Value must be an integer or integer string.",
+            "Value must be an integer or a string containing an integer.",
+            1,
+            id="disallow floats in range",
+        ),
+        pytest.param(
+            {
+                "name": "foo",
+                "type": "CHUNK[INT]",
+                "range": [1, 2],
```
leongdl, line 318: nit, do we need a test for range: [0] ? With the new `ge` code paths in the validators above.

leongdl (dismissed):

src/openjd/model/_internal/_validator_functions.py
```diff
@@ -0,0 +1,117 @@
...
+            # In this case, we cannot validate now. We need to validate this later when the template
+            # is instantiated into a job and the expressions have been resolved.
+            return value
+    elif isinstance(value, (int, float, Decimal)) and not isinstance(value, bool):
+        try:
+            float_value = Decimal(str(value))
+        except InvalidOperation:
+            raise ValueError("Value must be floating point.")
+    else:
+        raise ValueError("Value must be a float or a string containing a float.")
```
jusiskin: Is there a reason we aren't using a consistent error message about what the value must be?

jusiskin (dismissed):

src/openjd/model/_internal/_validator_functions.py
mwiebe: No reason, I can tighten that up.

test/openjd/model/v2023_09/test_chunk_int_task_parameter_type.py
mwiebe: Sure, I can drop that in.

leongdl (approved):

jusiskin (approved):

PR #172, opened 2025-02-19
feat!: Implement step parameter space iteration for chunks

mwiebe:

To use task chunking as defined by https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md, clients of the library need to be able to iterate over a parameter space based on chunking.

* Modify StepParameterSpaceIterator to handle parameter spaces that include a CHUNK[INT] parameter. The iteration pattern depends on the values in the 'chunks' object of the parameter definition.
   * If targetRuntimeSeconds is 0 or unspecified, split the CHUNK[INT] dimension into chunks, and then iterate over the whole space with that selection of chunks.
   * If targetRuntimeSeconds is a positive integer, move the CHUNK[INT] dimension to the innermost loop, and dynamically select the chunk each time it advances. Let the user of the iterator modify its chunks_default_task_count to adapt the chunk size over time.
   * The StepParameterSpaceIterator object behaved more like a pseudo-container than an iterator. Change it to act as the iterator itself (returning 'self' from __iter__() instead of a new object).
   * Changed how the special case zero-dimensional iteration works to be a Node/Iterator pair just like the others, for a more uniform interface.
* Changes to IntRangeExpr:
   * Fix a bug in IntRangeExpr.from_list that wasn't handling the end of a linear sequence correctly.
   * Change the string representation of IntRangeExpr to use "," instead of "-" when there are just two values.
* Add and update unit tests for all of the modifications.

Clients such as the openjd CLI command will be able to use chunked iteration of parameter spaces to run job templates that include chunking.

Implemented unit tests for the new chunked iteration modes.

Updated docstrings about chunked iteration.

This change includes a few small changes to the public contract of IntRangeExpr and the StepParameterSpaceIterator. Most code won't depend on these specifics, but some could.

No

----


src/openjd/model/_step_param_space_iter.py
```diff
@@ -82,48 +106,84 @@ def __init__(self, *, space: Optional[StepParameterSpace]):
...
+                    if param.chunks is None:
+                        raise ValueError(
+                            f"CHUNK[INT] parameter '{name}' must have a chunks definition."
+                        )
+                    self._chunks_adaptive = (
+                        param.chunks.targetRuntimeSeconds is not None
+                        and int(param.chunks.targetRuntimeSeconds) > 0
+                    )
+                    self._chunks_default_task_count = int(param.chunks.defaultTaskCount)
+                    break
```
jusiskin, line 121: We break out of the loop after finding the first parameter of type `CHUNK[INT]` here which led me to wonder if we only allow a single parameter of type `CHUNK[INT]` in a step parameter space.

After reviewing the RFC and OpenJobDescription specification, it was unclear that only one such parameter is allowed. Discussed with @mwiebe  offline and we agree that we need to amend the RFC, specification, and the model validation to model this assumption.

The plan is to add the validation for a single `CHUNK[INT]` parameter for a step parameter space in this PR and then back-fill the specification and RFC changes.

jusiskin (changes requested):
Great job with this code change.

I haven't finished my review pass yet, but wanted to surface one thing we should add to help clarify an assumption here.

src/openjd/model/_step_param_space_iter.py
mwiebe: I've added validation and a test for ensuring a single `CHUNK[INT]` parameter.

src/openjd/model/v2023_09/__init__.py
```diff
@@ -75,6 +75,7 @@
     StringRangeList,
     StringTaskParameterDefinition,
     TaskChunksDefinition,
+    TaskChunksRangeConstraint,
```
leongdl, line 78: Just curious, since we are adding chunks, is the model version changing too on release?

leongdl (commented):
This will take me some time to understand all the logic.

src/openjd/model/_step_param_space_iter.py
```diff
@@ -82,48 +106,84 @@ def __init__(self, *, space: Optional[StepParameterSpace]):
...
+        """
+        if not self._chunks_adaptive:
+            raise ValueError(
+                "The parameter space does not use adaptive chunking, so cannot modify chunks_default_task_count."
+            )
 
-            def __iter__(self):  # pragma: no cover
-                return self
+        self._expr_tree.set_chunks_default_task_count(value)
+        self._chunks_default_task_count = value
```
jusiskin: Should there be any input validation of the supplied value here? I'm thinking we want to at least confirm this is a non-negative int or `None`.

src/openjd/model/v2023_09/__init__.py
mwiebe: You can read https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0002-model-extensions.md to see the approach we worked out. The new chunking support is an extension to the 2023-09 model version, so you can use it with implementations supporting that extension. We'll wait before releasing a new model revision that incorporates the extensions until we've got a few more RFCs that went through the process.

src/openjd/model/_step_param_space_iter.py
mwiebe: We've merged a spec clarification here: https://github.com/OpenJobDescription/openjd-specifications/pull/73

src/openjd/model/v2023_09/__init__.py
leongdl: Thanks!

src/openjd/model/_step_param_space_iter.py
```diff
@@ -171,6 +320,84 @@ def __eq__(self, other: object) -> bool:
         )
 
 
+def divide_chunk_sizes(chunks_default_task_count: int, task_count: int) -> list[int]:
+    """Determines evenly distributed chunk lengths for a specified task count divided
+    into the requested number of chunks."""
+    if chunks_default_task_count <= 0:
+        chunk_count = 1
+    else:
+        chunk_count = (task_count + chunks_default_task_count - 1) // chunks_default_task_count
```
leongdl, line 331: That's an interesting way to implement something almost like ceiling!

src/openjd/model/_step_param_space_iter.py
```diff
@@ -134,17 +194,78 @@ def __getitem__(self, index: int) -> TaskParameterSet:
...
+                    )
+                if chunks_default_task_count is None:
+                    raise ValueError(
+                        f"CHUNK[INT] parameter '{name}' must have a default task count."
+                    )
+                # Expand the range to a list
+                if isinstance(parameter.range, list):
+                    parameter_range: list[int] = [int(v) for v in parameter.range]
+                else:
+                    parameter_range = list[int](IntRangeExpr.from_str(parameter.range))
```
jusiskin, line 230: Could we use a [generator expression](https://docs.python.org/3/reference/expressions.html#generator-expressions) instead here? Or does that cause problems for re-iterating the expression tree?

src/openjd/model/_step_param_space_iter.py
```diff
@@ -460,3 +745,128 @@ def __getitem__(self, index: int) -> TaskParameterSet:
...
+
+    def reset_iter(self) -> None:
+        self._index = 0
+
+    def next(self, result: TaskParameterSet) -> None:
+        if self._index >= len(self._node.range):
+            raise StopIteration()
+
+        chunks_default_task_count = self._node.chunks_default_task_count
+        range = self._node.range
```
jusiskin: nit: this shadows the built-in `range` function. Can we rename it to `parameter_range`?

jusiskin (changes requested):
Great job with this, @mwiebe! Just a few suggested changes here.

src/openjd/model/_step_param_space_iter.py
mwiebe: Yes, I'll add that. Also realized that comment was wrong, and it should always be a positive integer for this setter.

src/openjd/model/_step_param_space_iter.py
mwiebe: Using a generator expression wouldn't work, because some of the usage, like in `AdaptiveContiguousChunkIdentifierNodeIterator`, indexes into this list directly. We could use the `IntRangeExpr` object directly, but its indexing operator is not as efficient and I think we'd want a `next` implementation that used internal details of that object. So, I'd prefer to leave it as a list, and we can look at this kind of micro-optimization some other time.

src/openjd/model/_step_param_space_iter.py
mwiebe: Will do.

leongdl (approved):
I generally understand the logic, but probably need to do a few more thinking passes to walk through the logic.

src/openjd/model/_step_param_space_iter.py
mwiebe: If you messed around with graphics code like implementing https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm#Algorithm_for_integer_arithmetic and similar this stuff is pretty natural :)

src/openjd/model/_step_param_space_iter.py
jusiskin: Sounds good, thanks for the explanation

jusiskin (approved):

PR #177, opened 2025-02-28
feat!: Implement 'in' operator and chunksize overide for StepParamerSpaceIterator

mwiebe:

The OpenJD CLI lacks a way to validate whether a provided set of parameters are in the parameter space of the step they're for. When using chunks, adding this validation would make it easier to catch errors about whether the chunked parameter is contiguous and needs an interval like "3-3" vs a single value like "3".

With chunking, the step parameter space iterator doesn't have a mechanism, other than editing the input space object, to iterate over each individual task instead of over chunks of tasks. A mechanism like this is useful for scheduler integrations that reason at the task level and do their own chunking.

* The in operator and validate_containment function of the StepParameterSpaceIterator will allow the openjd-cli to validate that input task parameters are within the parameter space of steps that it is running. The latter
   * The purpose of this change is to support using chunked iteration from the OpenJD CLI. This validation helps ensure the `openjd run` command only accepts correct chunk, in addition to adding validation that was marked as a TODO in the openjd-cli codebase. Having validate_containment raise an exception with a message about why the task parameters are not in the parameter space results in better error messages.
* The chunksize override lets a caller iterate over every task of a parameter space regardless of the space's chunk size or adaptivity. This is useful for callers to evaluate the whole task space so that they can perform their own chunking logic later.
* Add a property chunks_parameter_name to the StepParameterSpaceIterator so that code using it can easily access the chunked parameter and its properties without having to perform its own loop over all of them.
* Use an IntRangeExpr instead of a string to hold the range expression of a RangeExpressionTaskParameterDefinition, the class used to instantiate a parameter definition template into a job.
* Found that negative steps in IntRange have bugs in corner cases. Code that is processing the list of IntRanges includes assumptions that the step is positive. To fix this, chose to normalize the step to be positive instead of adjusting the code to handle persistence of negative steps, as that is much simpler to get right.
   * Note that the IntRangeExpr was already not preserving the input order of a range expression, because it sorted all the components.
* Removed _BaseMessageError and used ValueError instead as a base class for the exceptions. This reduces custom code and interoperates with Pydantic better.

With a follow-up OpenJD CLI change, more pleasant CLI usage.

Added unit tests for the new functions and operator.

Updated docstrings

BREAKING CHANGE: The IntRangeExpr class now normalizes the steps of
    individual range components like "3-1:-2" to be positive like "1-3:2",
    so anything depending on the prior behavior needs to be updated.

- Does the change need to be threat modeled? For example, does it create or modify files/directories that must only be readable by the process owner?
    - If so, then please label this pull request with the "security" label. We'll work with you to analyze the threats.

No

----

### DIFF
--- a/src/openjd/model/_range_expr.py
+++ b/src/openjd/model/_range_expr.py
@@ -2,11 +2,14 @@
 
 from __future__ import annotations
 
-from bisect import bisect
+from bisect import bisect, bisect_left
 from collections.abc import Iterator, Sized
-from functools import total_ordering
 from itertools import chain
-from typing import Tuple
+from typing import Any, Tuple
+
+from pydantic import GetCoreSchemaHandler, GetJsonSchemaHandler
+from pydantic.json_schema import JsonSchemaValue
+from pydantic_core import core_schema
 
 from ._errors import ExpressionError, TokenError
 from ._tokenstream import Token, TokenStream, TokenType
@@ -15,15 +18,17 @@ from ._tokenstream import Token, TokenStream, TokenType
 class IntRangeExpr(Sized):
     """An Int Range Expression is a set of integer values represented as a sorted list of IntRange objects."""
 
-    _start: int
-    _end: int
+    _starts: list[int]
+    _ends: list[int]
     _ranges: list[IntRange]
     _length: int
     _range_length_indicies: list[int]
 
     def __init__(self, ranges: list[IntRange]):
+        if len(ranges) <= 0:
+            raise ExpressionError("Range expression cannot be empty")
         # Sort the ranges, then combine them where possible
-        sorted_ranges = sorted(ranges)
+        sorted_ranges = sorted(ranges, key=lambda v: (v._start, v._end, v._step))
         self._ranges = [sorted_ranges[0]]
         for range in sorted_ranges[1:]:
             if (
@@ -33,8 +38,8 @@ class IntRangeExpr(Sized):
                 self._ranges[-1] = IntRange(self._ranges[-1].start, range.end, range.step)
             else:
                 self._ranges.append(range)
-        self._start = self.ranges[0].start
-        self._end = self.ranges[-1].end
+        self._starts = [v.start for v in self.ranges]
+        self._ends = [v.end for v in self.ranges]
 
         # used to binary search ranges for __getitem__
         # ie. [32, 100, 132]
@@ -95,7 +100,7 @@ class IntRangeExpr(Sized):
         return ",".join(str(range) for range in self.ranges)
 
     def __repr__(self) -> str:
-        return f"{type(self).__name__}({self.ranges})"
+        return f"{type(self).__name__}.from_str({str(self)})"
 
     def __iter__(self) -> Iterator[int]:
         return chain(*self.ranges)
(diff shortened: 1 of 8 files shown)


src/openjd/model/_step_param_space_iter.py
```diff
@@ -699,6 +809,41 @@ def __len__(self) -> int:
...
+                raise ValueError(
+                    f"Parameter {self.name} of type {param.type.name} value {param.value} is not a contiguous interval like '1-5' as required by the chunk range constraint."
+                )
+            # The chunk must be a subset of the range
+            try:
+                for v in IntRangeExpr.from_str(param.value):
+                    if v not in self.range_set:
+                        raise ValueError(
+                            f"Parameter {self.name} of type {param.type.name} value {param.value} is not a valid range expression like '1-5,10' as required by the chunk range constraint."
+                        )
```
epmog: I believe this error message has to be updated to indicate it's not a proper subset, rather than a valid range expression

src/openjd/model/_step_param_space_iter.py
```diff
@@ -872,3 +1062,23 @@ def iter(self) -> AdaptiveNoncontiguousChunkIdentifierNodeIterator:
 
     def set_chunks_default_task_count(self, value: int) -> None:
         self.chunks_default_task_count = value
+
+    def validate_containment(self, params: TaskParameterSet):
+        """Checks if the params restricted to this node are part of the node's range."""
+        param = params[self.name]
+        if param.type != self.type:
+            raise ValueError(
+                f"Parameter {self.name} of type {param.type.name} value {param.value} is not in the parameter space range."
```
epmog: Should this error message be about `self.type` and `param.type` not matching?

src/openjd/model/v2023_09/_model.py
```diff
@@ -545,22 +545,10 @@ def coerce(v: Any) -> Any:
 class RangeExpressionTaskParameterDefinition(OpenJDModel_v2023_09):
     # element type of items in the range
     type: TaskParameterType
-    range: TaskRangeExpression
+    range: IntRangeExpr
```
epmog, line 547: Is `TaskRangeExpression` then deprecated in the model and will be removed in a future revision? It does not appear to be used anywhere else. This change probably also needs to be called out in the breaking changes.

I guess we can do this change, because everywhere else uses `RangeString` to allow for format strings and all the usage here for format strings is already resolved.

epmog (commented):
Looks good! Just the few comments about errors and the changing type

src/openjd/model/_step_param_space_iter.py
mwiebe: Yes, will fix.

src/openjd/model/_step_param_space_iter.py
mwiebe: Yes, that will be better.

src/openjd/model/v2023_09/_model.py
mwiebe: Before this change, `TaskRangeExpression` is only being used implicitly in the model, it isn't a type that's provided in the public interface. I think also deleting the alias to `RangeString` makes sense.

epmog (dismissed):

epmog (dismissed):

src/openjd/model/_range_expr.py
```diff
@@ -156,31 +165,58 @@ def _validate(self) -> None:
...
-        elif self._step < 0:
-            offset = -1
-
-        self._range = range(self._start, self._end + offset, self._step)
-
-        self._validate()
+        if step > 0:
+            if start > end:
+                raise ValueError("Range: a descending range must have a negative step")
+            self._range = range(start, end + 1, step)
```
crowecawcaw, line 206: Should this be `range(start, end + step, step)` in case the step is not 1 but we still want to include the end value? Similar in the negative branch below. Or if this does work as expected, is there a test case that would verify the final value is included in a range with a different step size?

mwiebe: Consider these cases:

```
>>> "1-5:3" # 1 to 5 (inclusive), stepping by 3. Values [1, 4]
'1-5:3'
>>> list(range(1,5 + 1,3)) # With + 1
[1, 4]
>>> list(range(1,5 + 3,3)) # With + step
[1, 4, 7]
```

It wasn't obvious whether this was being tested thoroughly, so I dropped in a few more test cases.

crowecawcaw (approved):

src/openjd/model/_range_expr.py
crowecawcaw: Ah, I had a bad assumption that ranges could include floats. Looks like that's not true of OJD or Python's `range`.

epmog (approved):

PR #182, opened 2025-03-24
refactor!: Update FormatString expression to accept parsing context

mwiebe:

To support Open Job Description modeling parsing that differs depending on the revision and/or enabled extensions, the code needs a way to parse differently based on those values. This is required in order to experiment with proof of concepts for https://github.com/OpenJobDescription/openjd-specifications/discussions/79.

The question is, where to split the processing, should we do it inside the FormatString code, or should we do it in the models that use the FormatString code. This PR chooses the former because making the code choose the behavior in FormatString will be simpler to experiment with and will decouple expression-related changes from the rest of the Open Job Description models.

Refactor the expression string processing to make putting together different proofs of concept for https://github.com/OpenJobDescription/openjd-specifications/discussions/79 easier.

* Introduce an abstract base class, ModelParsingContextInterface, that every model parsing context must be a subclass of. It has variables spec_rev and extensions, for parsing to reference.
* Modify the DynamicConstrainedStr class to require a ModelParsingContextInterface during construction. The FormatString to handle template the expression syntax is a subset, so modifying this class to handle the Pydantic integration was the simplest approach.
   * As part of this, change DynamicConstrainedStr and subclasses to use __new__ instead of __init__, so they all handle construction uniformly with the required context parameter.
* Create a function format_string_expr_parse that handles parsing a string according to the format string expression parsing syntax, using a model parsing context to determine the specific syntax. This does not change the syntax, only introduces a place to add new syntax support.
* Update the template variable reference pre-validation and template -> job instantiation code to work with the updated DynamicConstrainedStr that now requires a context. This required the instantiation logic to be more clear about the separation between Template and Job, since the latter instantiation happens without a context.

Putting together different proofs of concept for https://github.com/OpenJobDescription/openjd-specifications/discussions/79 will be easier.

After this change, the code to use Python `ast` for the expression parsing as considered in https://github.com/OpenJobDescription/openjd-specifications/discussions/79#discussioncomment-12467858 is quite small: https://github.com/mwiebe/openjd-model-for-python/commit/4456628e57ddd277235e37fcb031a47339561096.

* Everywhere that the tests do model validation or construct DynamicConstrainedStr/FormatString instances, introduce the model parsing context. In some places had to change the content of the tests to reflect the more strict separation between model parsing and job instantiation.
* Change various different model construction used in the tests to call decode_job_template instead. The latter function has its own tests, and using the specified Open Job Description syntax instead of object constructors isolates the tests from implementation details.

Docstrings for affected classes were updated.

BREAKING CHANGE: Any creation of a DynamicConstrainedStr or FormatString
    will need to provide a model parsing context, that includes the Open
    Job Description revision and any extensions that are enabled.

No

----

### DIFF
--- a/src/openjd/model/_internal/_create_job.py
+++ b/src/openjd/model/_internal/_create_job.py
@@ -1,22 +1,56 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-from typing import cast, Any, Union
+from contextlib import contextmanager
+from typing import Annotated, Any, Union
 
 from pydantic import ValidationError
+from pydantic import TypeAdapter
 from pydantic_core import InitErrorDetails
 
 from .._symbol_table import SymbolTable
-from .._format_strings import FormatString, FormatStringError
+from .._format_strings import FormatString
 from .._types import OpenJDModel
 
 __all__ = ("instantiate_model",)
 
 
+@contextmanager
+def capture_validation_errors(
+    *, output_errors: list[InitErrorDetails], loc: tuple[Union[str, int], ...], input: Any
+):
+    """Context manager to collect validation errors from pydantic into a list.
+
+    Args:
+        output_error_list (list[InitErrorDetails]): The list to collect errors into.
+        loc (tuple): The location of the input value being validated.
+        input (Any): The input value being validated.
+    """
+    try:
+        yield
+    except ValidationError as exc:
+        # Convert the ErrorDetails to InitErrorDetails by extending the 'loc' and excluding the 'msg'
+        for error_details in exc.errors():
+            init_error_details: dict[str, Any] = {}
+            for err_key, err_value in error_details.items():
+                if err_key == "loc":
+                    init_error_details["loc"] = (*loc, *err_value)  # type: ignore
+                elif err_key != "msg":
+                    init_error_details[err_key] = err_value
+            output_errors.append(init_error_details)  # type: ignore
+    except ValueError as exc:
+        output_errors.append(
+            InitErrorDetails(
+                type="value_error",
+                loc=loc,
+                ctx={"error": exc},
+                input=input,
+            )
+        )
+
+
 def instantiate_model(  # noqa: C901
     model: OpenJDModel,
     symtab: SymbolTable,
-    loc: tuple[Union[str, int], ...] = tuple[str](),
-    within_field: str = "",
 ) -> OpenJDModel:
     """This function is for instantiating a Template model into a Job model.
 
@@ -27,9 +61,6 @@ def instantiate_model(  # noqa: C901
         model (OpenJDModel): The model instance to transform.
         symtab (SymbolTable): The symbol table containing fully qualified Job parameter values
             used during the instantiation.
-        loc (tuple[Union[str,int], ...], optional): Path to the model. Used for generating contextual errors.
-        within_field (str): The name of the field where this model was found during traversal. "" if this
-            is the top-level/root model
 
     Raises:
         ValidationError - If there are any validation errors from the target models.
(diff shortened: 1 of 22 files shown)


src/openjd/model/_internal/_create_job.py
```diff
@@ -40,50 +41,87 @@ def instantiate_model(  # noqa: C901
...
+                InitErrorDetails(
+                    type="value_error",
+                    loc=(*loc, field_name),
+                    ctx={"error": exc},
+                    input=field_value,
+                )
+            )
+
+        try:
+            if instantiated is not None:
```
crowecawcaw: Could the instantiated value have a legitimate value of `None`?

src/openjd/model/_internal/_create_job.py
```diff
@@ -225,11 +265,20 @@ def _instantiate_list_field(  # noqa: C901
...
                     )
                 )
+            except ValueError as exc:
+                errors.append(
+                    InitErrorDetails(
+                        type="value_error",
+                        loc=(*loc, field_name, idx),
+                        ctx={"error": exc},
+                        input=item,
+                    )
```
crowecawcaw: Would it make sense to abstract this code? It's repeated a few times.

test/openjd/model/v2023_09/test_template_variables.py
```diff
@@ -43,1110 +43,1104 @@ def make_script(env_or_task: str, scriptname: str, scriptdata: str) -> dict:
...
+    pytest.param(
+        {
+            "specificationVersion": "jobtemplate-2023-09",
+            "name": "Foo {{Param.Foo}}",
+            "parameterDefinitions": [FOO_PARAMETER_INT],
+            "steps": [STEP_TEMPLATE_FOO],
+        },
+        id="minimum int parameter",
+    ),
+    pytest.param(
```
crowecawcaw, line 56: The diff on this test file isn't the easiest to scan. Is this just a formatting change? Do the original tests still pass?

mwiebe: I moved the test cases list to a separate variable because it improves the output significantly and speeds up debugging iterations. This part has changes just like other tests. I'll see what I can do to extract a better diff.

mwiebe: After re-indenting a few levels and applying auto-formatting for the narrower space, here's the diff that results for this file.

```diff
diff --git a/test/openjd/model/v2023_09/test_template_variables.py b/test/openjd/model/v2023_09/test_template_variables.py
index 40f89c6..2009b81 100644
--- a/test/openjd/model/v2023_09/test_template_variables.py
+++ b/test/openjd/model/v2023_09/test_template_variables.py
@@ -6,7 +6,7 @@ import pytest
 from pydantic import ValidationError

 from openjd.model._parse import _parse_model
-from openjd.model.v2023_09 import JobTemplate
+from openjd.model.v2023_09 import JobTemplate, ModelParsingContext as ModelParsingContext_v2023_09

 # Some minimal data to reference in tests.
 ENV_SCRIPT_FOO = {
@@ -44,8 +44,7 @@ def make_script(env_or_task: str, scriptname: str, scriptdata: str) -> dict:


 class TestJobTemplateVars:
-    @pytest.mark.parametrize(
-        "data",
+    PARAMETRIZE_CASES: tuple = (
         (
             pytest.param(
                 {
@@ -482,7 +481,12 @@ class TestJobTemplateVars:
             ),
         ),
     )
-    def test_parse_success(self, data: dict[str, Any]) -> None:
+
+    @pytest.mark.parametrize(
+        "data",
+        PARAMETRIZE_CASES,
+    )
+    def test_job_template_variables_parse_success(data: dict[str, Any]) -> None:
         # Parsing tests of valid Open Job Description JobTemplate
         # It is sufficient to check that parsing the input does not
         # raise an exception. We trust the Pydantic package's testing
@@ -490,13 +494,12 @@ class TestJobTemplateVars:
         # constructed for valid input.

         # WHEN
-        _parse_model(model=JobTemplate, obj=data)
+        _parse_model(model=JobTemplate, obj=data, context=ModelParsingContext_v2023_09())

         # THEN
         # no exception was raised.

-    @pytest.mark.parametrize(
-        "data,error_count",
+    PARAMETRIZE_CASES = (
         (
             pytest.param(
                 {
@@ -1141,12 +1144,17 @@ class TestJobTemplateVars:
             ),
         ),
     )
-    def test_parse_fails(self, data: dict[str, Any], error_count: int) -> None:
+
+    @pytest.mark.parametrize(
+        "data,error_count",
+        PARAMETRIZE_CASES,
+    )
+    def test_job_template_variables_parse_fails(data: dict[str, Any], error_count: int) -> None:
         # Failure case testing for Open Job Description JobTemplate.

         # WHEN
         with pytest.raises(ValidationError) as excinfo:
-            _parse_model(model=JobTemplate, obj=data)
+            _parse_model(model=JobTemplate, obj=data, context=ModelParsingContext_v2023_09())

         # THEN
         assert len(excinfo.value.errors()) == error_count, str(excinfo.value)
```

src/openjd/model/_internal/_create_job.py
mwiebe: Yes, if one of the calls to `_instantiate_*` raises an exception, then the value will be `None` as it is not modified from line 68.

src/openjd/model/_internal/_create_job.py
mwiebe: Yes, I think a context manager could work well.

test/openjd/model/v2023_09/test_template_variables.py
crowecawcaw: Thank you!

crowecawcaw (dismissed):

src/openjd/model/_internal/_create_job.py
mwiebe: I did some refactoring and found this was a significant simplification.

crowecawcaw (approved):

src/openjd/model/_internal/_create_job.py
```diff
@@ -1,22 +1,56 @@
...
+    try:
+        yield
+    except ValidationError as exc:
+        # Convert the ErrorDetails to InitErrorDetails by extending the 'loc' and excluding the 'msg'
+        for error_details in exc.errors():
+            init_error_details: dict[str, Any] = {}
+            for err_key, err_value in error_details.items():
+                if err_key == "loc":
+                    init_error_details["loc"] = (*loc, *err_value)  # type: ignore
+                elif err_key != "msg":
```
leongdl, line 37: nit, do we care about any other else? Like future "unknown" errors in case they get swallowed ?

leongdl: From the docs, eg "type" is also a key type.

mwiebe: The InitErrorDetails class https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.InitErrorDetails is well-defined. The reason the code is filtering it out because ErrorDetails has that field but InitErrorDetails does not.

leongdl (approved):

PR #184, opened 2025-04-23
fix: improve parameter definition validation and error reporting

### LINKED ISSUE #179: Bug: Parameter validation fails when allowedValues is set to None
### Describe the bug

The parameter definition validators fail with `'NoneType' object is not iterable` when `allowedValues` is explicitly set to `None`, despite `None `being a valid value according to the model definition (`Optional[AllowedParameterStringValueList] = None`).

### Expected Behaviour

The validation should succeed in both cases:
1. When allowedValues is omitted (implicit None)
2. When allowedValues is explicitly set to None

### Current Behaviour

Explicitly setting it to None causes a validation error:
`'NoneType' object is not iterable` when trying to validate the parameter definition.

### Reproduction Steps

1. Create a parameter definition with allowedValues explicitly set to None:
```python
{
    "name": "Foo",
    "type": "STRING",
    "allowedValues": None
}
```

2. Attempt to validate this definition using any of the parameter definition classes (JobStringParameterDefinition, JobPathParameterDefinition, JobIntParameterDefinition, or JobFloatParameterDefinition)

3. Observe the validation error

### Environment

The issue should be reproducible in any environment as it's a logic error in the validation code.

ryyakobe:
# Fix parameter definition validation and error reporting

This PR improves parameter definition validation across all parameter types to properly handle `allowedValues` and provide clearer error messages.

Fixes: https://github.com/OpenJobDescription/openjd-model-for-python/issues/179

There were several issues with parameter definition validation:
1. The validators were rejecting explicit `None` values for `allowedValues`, without a clear error message
2. Error location reporting was inconsistent in validation error messages
3. Documentation comments for `maxValue` parameters incorrectly described them as "Minimum value"
4. Type casting in `JobPathParameterDefinition` was using the wrong class name

1. Added explicit validation for `None` values in `allowedValues` across all parameter types (String, Path, Int, Float) with clear error messages
2. Fixed error location reporting in validation error messages to use consistent `("allowedValues", i)` format
3. Corrected documentation comments for `maxValue` parameters to say "Maximum value"
4. Fixed type casting in `JobPathParameterDefinition` to use the correct class name
5. Added test cases for both implicit and explicit `None` handling

This change ensures that:
1. Explicitly set `None` values for `allowedValues` are properly rejected with clear error messages
2. Implicitly omitted `allowedValues` continue to work correctly
3. Error messages are more consistent and accurate across all parameter types
4. Documentation comments correctly describe parameter constraints

Added test cases for both implicit and explicit `None` handling across all parameter types. All existing and new tests pass successfully.

Yes, all unit tests pass.

Yes, fixed incorrect documentation comments for `maxValue` parameters.

No, this is not a breaking change. It improves validation to reject invalid inputs that should have been rejected before, while maintaining compatibility with valid inputs. The error message itself has changed, which could be considered a breaking change but the benefit should be worth it. Compare:
```
'NoneType' object is not iterable
```
to
```
allowedValues cannot be None. The field must contain at least one value or be omitted entirely.
```

- Does the change need to be threat modeled?
No.

----

### DIFF
--- a/src/openjd/model/v2023_09/_model.py
+++ b/src/openjd/model/v2023_09/_model.py
@@ -56,6 +56,11 @@ from .._types import (
     TemplateVariableDef,
 )
 
+# Error message constants
+_ALLOWED_VALUES_NONE_ERROR = "allowedValues cannot be None. The field must contain at least one value or be omitted entirely."
+_VALUE_LESS_THAN_MIN_ERROR = "Value less than minValue."
+_VALUE_LARGER_THAN_MAX_ERROR = "Value larger than maxValue."
+
 
 class ModelParsingContext(ModelParsingContextInterface):
     """Context required while parsing an OpenJDModel. An instance of this class
@@ -1211,6 +1216,8 @@ class JobStringParameterDefinition(OpenJDModel_v2023_09, JobParameterInterface):
     def _validate_allowed_values_item(
         cls, value: AllowedParameterStringValueList, info: ValidationInfo
     ) -> AllowedParameterStringValueList:
+        if value is None:
+            raise ValueError(_ALLOWED_VALUES_NONE_ERROR)
         min_length = info.data.get("minLength")
         max_length = info.data.get("maxLength")
         errors = list[InitErrorDetails]()
@@ -1421,7 +1428,7 @@ class JobPathParameterDefinition(OpenJDModel_v2023_09, JobParameterInterface):
             "default",
         },
         adds_fields=lambda this, symtab: {
-            "value": symtab[f"RawParam.{cast(JobStringParameterDefinition,this).name}"]
+            "value": symtab[f"RawParam.{cast(JobPathParameterDefinition,this).name}"]
         },
     )
 
@@ -1451,8 +1458,10 @@ class JobPathParameterDefinition(OpenJDModel_v2023_09, JobParameterInterface):
     @field_validator("allowedValues")
     @classmethod
     def _validate_allowed_values_item(
-        cls, value: ParameterStringValue, info: ValidationInfo
-    ) -> ParameterStringValue:
+        cls, value: AllowedParameterStringValueList, info: ValidationInfo
+    ) -> AllowedParameterStringValueList:
+        if value is None:
+            raise ValueError(_ALLOWED_VALUES_NONE_ERROR)
         min_length = info.data.get("minLength")
         max_length = info.data.get("maxLength")
         errors = list[InitErrorDetails]()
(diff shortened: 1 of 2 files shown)


src/openjd/model/v2023_09/_model.py
```diff
@@ -1196,7 +1200,7 @@ def _validate_allowed_values_item(
                     errors.append(
                         InitErrorDetails(
                             type="value_error",
-                            loc=(i,),
+                            loc=("allowedValues", i),
```
mwiebe: This looks like a fix to a bug where the breadcrumbs were wrong. There are a few tests that assert on these breadcrumbs, like https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/test/openjd/model/test_create_job.py#L816 and https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/test/openjd/model/test_parse.py#L259. Could you add a similar assertion about this?

src/openjd/model/v2023_09/_model.py
```diff
@@ -1438,7 +1446,7 @@ def _validate_allowed_values_item(
                     errors.append(
                         InitErrorDetails(
                             type="value_error",
-                            loc=("allowedValues", i),
```
ryyakobe, line 1465: Before, the error included `allowedValues` twice for `JobPathParameterDefinition`:
```
'loc': ('allowedValues', 'allowedValues', 2)
```
After:
```
('allowedValues', 2)
```

This is now also consistent with `JobStringParameterDefinition` and other job parameter types as well.

src/openjd/model/v2023_09/_model.py
ryyakobe: Great catch!

mwiebe (dismissed):

test/openjd/model/v2023_09/test_job_parameters.py
```diff
@@ -152,6 +152,13 @@ class TestJobStringParameterDefinition:
                 },
                 id="all fields",
             ),
+            pytest.param(
+                {
+                    "name": "Foo",
+                    "type": "STRING",
+                },
+                id="allowedValues is implicitly None",
```
baxeaz: Omitting we intend to be legal which makes sense - but we consider that "implicit None" rather than "implicit Any" or similar?  In our code we test for None as an error case, so None even implicitly seems odd that our tests are saying it should be valid

ryyakobe: Would renaming this id to `allowedValues not provided` address your comment?

baxeaz: Yep that would work!

mwiebe (dismissed):

test/openjd/model/v2023_09/test_job_parameters.py
ryyakobe: Renamed.

baxeaz (dismissed):

mwiebe (approved):

baxeaz (approved):

PR #200, opened 2025-06-06
feat: Exposing ExtensionName publicy

baxeaz:
We want to expose the ExtensionName publicly for consumers of openjd-model to use

Expose it

Can access the enum

hatch run test, used in changes I'm making to the Unreal integration

No

No

No

----


crowecawcaw (approved):

jusiskin (changes requested):
This should be a `feat:` and not a `fix:`, right?

baxeaz:
> This should be a `feat:` and not a `fix:`, right?

Hmm I guess?  Updated!

jusiskin (approved):

mwiebe (approved):

PR #206, opened 2025-06-16
chore: add allow_zero_version to pyproject.toml in response to semant…

evanspearman-a:
What was the problem/requirement? (What/Why)

Semantic release broke a default setting so it would force our project to go version 1.0.0.
What was the solution? (How)

Add a flag in the pyproject.toml to re-enable zero versions.
What is the impact of this change?

Gets back to our version structure
How was this change tested?

N/A
Was this change documented?

No.
Is this a breaking change?
No.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -149,6 +149,7 @@ fail_under = 94
 # Can be removed or set to true once we are v1
 major_on_zero = false
 tag_format = "{version}"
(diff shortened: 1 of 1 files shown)


AWS-Samuel (approved):

mwiebe (approved):

miabatta:
Fixed in #209

PR #212, opened 2025-07-04
fix: Adding a TypeAdapter cache to fix a performance regression with …

baxeaz:
Recent changes to type validation have created a performance bottleneck with larger templates in pydantic's TypeAdapter.

- Add a TypeAdapter cache following [pydantic performance guidance](https://docs.pydantic.dev/latest/concepts/performance/#typeadapter-instantiated-once)
- Add a benchmark test to help diagnose and debug performance bottlenecks

~87% performance improvement when testing with larger templates

- All unit tests passed
- New benchmark test added which showed a ~87% performance improvement with vs without the cache.
- End to End testing performed in deadline cloud with the patch applied showed the fix resolved the failing test which originally uncovered the performance issue

Yes

No

No

----

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -5,7 +5,8 @@ pre-install-commands = [
 
 [envs.default.scripts]
 sync = "pip install -r requirements-testing.txt"
-test = "pytest --cov-config pyproject.toml {args:test}"
+test = "pytest test/ --cov-config pyproject.toml --ignore=test/openjd/model/benchmark {args}"
+benchmark = "pytest test/openjd/model/benchmark --no-cov {args}"
 typing = "mypy {args:src test}"
 style = [
   "ruff check {args:src test}",
--- a/src/openjd/model/_internal/_create_job.py
+++ b/src/openjd/model/_internal/_create_job.py
@@ -1,7 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
 from contextlib import contextmanager
-from typing import Annotated, Any, Union
+from typing import Annotated, Any, Union, Dict
 
 from pydantic import ValidationError
 from pydantic import TypeAdapter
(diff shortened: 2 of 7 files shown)


crowecawcaw (dismissed):

src/openjd/model/_internal/_create_job.py
```diff
@@ -13,6 +13,25 @@
...
+    """Get a TypeAdapter for the given field type, using a cache for efficiency.
+
+    Args:
+        field_type: The type to adapt.
+
+    Returns:
+        A TypeAdapter for the given type.
+    """
+    # Use the string representation of the type as a cache key
+    type_key = str(field_type)
```
mwiebe: Did you try `id` as an alternative to `str` as well? If types aren't too redundant, it might be a bit faster.

mwiebe (dismissed):

test/openjd/model/benchmark/test_benchmark_step_environments.py
```diff
@@ -0,0 +1,115 @@
...
+            ps = pstats.Stats(profiler, stream=s).sort_stats("time")
+            ps.print_stats(20)
+            logger.info("TOP 20 FUNCTIONS BY TOTAL TIME:")
+            for line in s.getvalue().splitlines():
+                logger.info(line)
+
+            # Verify that the operation completed within a reasonable time
+            assert (
+                elapsed_time < 10
+            ), f"Operation took {elapsed_time:.2f} seconds, which exceeds the 10 second threshold"
```
epmog, line 96: How close are we to this threshold?

hatch.toml
```diff
@@ -5,7 +5,8 @@ pre-install-commands = [
 
 [envs.default.scripts]
 sync = "pip install -r requirements-testing.txt"
-test = "pytest --cov-config pyproject.toml {args:test}"
+test = "pytest test/ --cov-config pyproject.toml --ignore=test/openjd/model/benchmark {args}"
+benchmark = "pytest test/openjd/model/benchmark --no-cov {args}"
```
epmog, line 9: Should we be running these in GitHub actions to prevent regression before merge?

src/openjd/model/_internal/_create_job.py
baxeaz: It was indeed an improvement!  I had some concerns about how those ids would resolve/collide but after doing a bit more investigation/testing looks like this is the way to go.  Thanks!

test/openjd/model/benchmark/test_benchmark_step_environments.py
baxeaz: It's going to depend on the host which is part of why I've relegated it to "benchmark", but I'm running it in about 4 seconds currently

hatch.toml
baxeaz: I was looking at that - I'd like to, but it looks like the commands run live in a shared config currently.  If there's a simple way to add that I'm happy to do it though!

mwiebe (approved):

crowecawcaw (approved):

PR #219, opened 2025-07-31
feat: Attempting to load use CSafeLoader for faster YAML parsing wher…

baxeaz:
Our yaml parser was far slower than our JSON parser which prevented certain checks we'd like to do at job creation time in some cases and generally caused our create job API to run slower than we'd like.

Try to use the yaml CSafeLoader where available.  Create benchmarking tests to note the difference.

```
### 1. Small Template (987 characters):
• **SafeLoader**: 2.91ms
• **CSafeLoader**: 0.27ms
• **Performance improvement**: 11.0x faster
• **Time saved**: 2.64ms per parse

### 2. Large Template (44,187 characters):
• **SafeLoader**: 125.74ms
• **CSafeLoader**: 13.44ms
• **Performance improvement**: 9.4x faster
• **Time saved**: 112.30ms per parse

### 3. Extra Large Template (192,387 characters):
• **SafeLoader**: 590.07ms
• **CSafeLoader**: 65.48ms
• **Performance improvement**: 9.0x faster
• **Time saved**: 524.59ms per parse

### 4. File-based Template (27,347 characters):
• **SafeLoader**: 77.26ms
• **CSafeLoader**: 8.08ms
• **Performance improvement**: 9.6x faster
• **Time saved**: 69.18ms per parse
```

All unit tests pass
Benchmarking test created which passes and shows above performance improvements

Yes

No

No - the CSafeLoader is a C implementation of the safe loader we were using previously - https://pyyaml.org/wiki/PyYAMLDocumentation

----

### DIFF
--- a/src/openjd/model/_parse.py
+++ b/src/openjd/model/_parse.py
@@ -1,6 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
 import json
+import os
 from dataclasses import is_dataclass
 from decimal import Decimal
 from enum import Enum
@@ -31,6 +32,18 @@ __all__ = (
 class DocumentType(str, Enum):
     JSON = "JSON"
     YAML = "YAML"
+
+
+# Environment variable can optionally disable CSafeLoader (used for benchmarking)
+use_csafe_loader = os.environ.get("OPENJD_USE_CSAFE_LOADER", "true").lower() in ("true", "1", "yes")
+
+try:
+    if use_csafe_loader:
+        from yaml import CSafeLoader as _YamlLoader  # type: ignore[attr-defined]
+    else:
+        raise ImportError("CSafeLoader disabled by environment variable")
+except ImportError:
+    from yaml import SafeLoader as _YamlLoader  # type: ignore[assignment]
 
 
 # Pydantic injects a __pydantic_model__ attribute into all dataclasses. To be able to parse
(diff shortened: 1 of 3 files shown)


src/openjd/model/_parse.py
```diff
@@ -111,7 +124,7 @@ def document_string_to_object(*, document: str, document_type: DocumentType) ->
         if document_type == DocumentType.JSON:
             parsed_document = json.loads(document)
         else:  # YAML
-            parsed_document = yaml.safe_load(document)
+            parsed_document = yaml.load(document, Loader=_YamlLoader)
```
epmog, line 127: Doubled checked cause I was interested. `safe_load(doc)` _is_ equivalent to `load(doc, SafeLoader)`: https://github.com/yaml/pyyaml/blob/69c141adcf805c5ebdc9ba519927642ee5c7f639/lib/yaml/__init__.py#L117-L125

src/openjd/model/_parse.py
```diff
@@ -33,6 +34,18 @@ class DocumentType(str, Enum):
...
 
 
+# Environment variable can optionally disable CSafeLoader (used for benchmarking)
+use_csafe_loader = os.environ.get("OPENJD_USE_CSAFE_LOADER", "true").lower() in ("true", "1", "yes")
+
+try:
+    if use_csafe_loader:
+        from yaml import CSafeLoader as _YamlLoader  # type: ignore[attr-defined]
+    else:
+        raise ImportError("CSafeLoader disabled by environment variable")
```
mwiebe, line 44: How does benchmarking run the non-C loader with this `raise` here?

mwiebe: Ah I see, I think I was sidetracked by the informative message here that looks like it has purpose to it. The exception from line 44 is always caught in line 45.

mwiebe (approved):

epmog (approved):

Issue #252, opened 2025-12-15
Bug: AttributeError when decoding environment template with environment: None


crowecawcaw:
### Describe the bug

decode_environment_template() raises an unexpected AttributeError instead of DecodeValidationError when the environment field is None.

### Reproduction

python
from openjd.model import decode_environment_template

decode_environment_template(template={
    'specificationVersion': 'environment-2023-09',
    'environment': None
})

### Expected Behaviour

Should raise DecodeValidationError indicating that environment cannot be None.

### Current Behaviour

Raises AttributeError: 'NoneType' object has no attribute 'get'

### Reproduction Steps

```python
from openjd.model import decode_environment_template

decode_environment_template(template={
    'specificationVersion': 'environment-2023-09',
    'environment': None
})```

### Environment

- openjd-model: 0.8.5
- Commit: 0fa2a3738d1b6b28e468f538c9d97e6b68574151
- Mac 
- Python 3.13.2

PR #253, opened 2025-12-15
test: add fuzzer

crowecawcaw:
There are gaps in the parser where a malformed template can result in an unexpected error.

Add a fuzzer.

It discovered the issue we were seeing in our code that depends on this parser: https://github.com/OpenJobDescription/openjd-model-for-python/issues/252

Ran it against this codebase and discovered a bug.

Yes, script is self-documenting.

No

Yes, improvement
----

### DIFF
--- a/test/openjd/model/test_fuzz.py
+++ b/test/openjd/model/test_fuzz.py
@@ -15,6 +15,7 @@ import string
 import sys
 import time
 import traceback
+from typing import Optional
 
 from openjd.model import (
     DecodeValidationError,
@@ -197,7 +198,7 @@ def fuzz_yaml_json_string():
     )
 
 
-def run_fuzzer(num_tests: int, seed: int | None, verbose: bool) -> bool:
+def run_fuzzer(num_tests: int, seed: Optional[int], verbose: bool) -> bool:
     if seed is not None:
         random.seed(seed)
     else:
--- a/test/openjd/model/test_fuzz.py
+++ b/test/openjd/model/test_fuzz.py
@@ -5,7 +5,8 @@ Fuzzer for openjd-model-for-python to ensure malformed templates
 only raise DecodeValidationError or NotImplementedError.
 
 Usage:
-    python scripts/fuzz.py [--num-tests N] [--seed S] [--verbose]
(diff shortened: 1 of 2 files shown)


prestomation:
Should this be part of the pytest suite?

crowecawcaw:
> Should this be part of the pytest suite?

I wasn't sure. A fuzzer might run for a long period of time if you do a lot of iterations. I'll enable it as a pytest as well with a smaller number of iterations that complete in a reasonable period of time.

AWS-Samuel (approved):
High level comment that there's some hard-coded `2023-09` strings we might want to parametrize. I don't think there'll be a new spec version in the near future so we can probably put this off

test/openjd/model/test_fuzz.py
```diff
@@ -0,0 +1,298 @@
...
+                doc = fuzz_yaml_json_string()
+                document_string_to_object(document=doc, document_type=DocumentType.JSON)
+            else:
+                doc = fuzz_yaml_json_string()
+                document_string_to_object(document=doc, document_type=DocumentType.YAML)
+
+            successes += 1
+
+        except EXPECTED_EXCEPTIONS:
+            successes += 1
```
jusiskin, line 234: There is one concern here. We expect some of these inputs to raise exceptions (e.g. `DecodeValidationError`) but others to not raise exceptions. But we do not have logic that to know which templates we expect to raise such errors.

So there is a chance that we are ignoring errors in our job/environment template decoder or allowing bad inputs through.

Is that something we are okay with? Are the goals of the test to only identify unhandled exceptions hidden in the decoder?

jusiskin (changes requested):

test/openjd/model/test_fuzz.py
crowecawcaw: Good point! My original goal was to just figure out why the library sometimes raises exceptions that aren't `DecodeValidationError` `NotImplementedError` because it broke some error handling code based on this library. Expanding this to test for correctness also makes sense. What do you think about leaving that for the future if we want to add it?

jusiskin: I'm fine with that since the repository already has existing correctness tests.

jusiskin (approved):

PR #258, opened 2026-01-30
feat!: Implement FEATURE_BUNDLE_1 RFC 0004

mwiebe:

[RFC 0004](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0004-enhanced-limits-and-capabilities.md) was accepted 1/9, and needs implementation in openjd-model, openjd-sessions, etc.

Implement the extension in the model parsing. For the syntax sugar, this also includes a new method on the StepTemplate that de-sugars it into the syntax without the sugar.

Progress towards this extension being available for use.

This change introduces unit tests for the extension.

I used the following prompt for LLM evaluation of this code change against RFC 0004 in openjd-specifications:

> You are a helpful RFC conformance reviewer. Another agent has
> implemented RFC 0004, along with tests, in the most recent commit. Your
> job is to carefully inspect the RFC, the RFC-specific details in the
> OpenJD spec, and look at all the code changes to evaluate compliance.
> Produce a report on the quality of the implementation. How well does it
> conform? Give each relevant section/subsection of the spec a score. Also
> look at the unit tests and evaluate that it is covering all edge cases
> of the spec. Write your report to the file CONFORMANCE_REPORT.md.

All code also reviewed by hand.

Extension documented in openjd-specifications, this change adds comments/docstrings as appropriate.

Yes, in a subtle way with some fields that can now be FormatString instead of just int.

BREAKING CHANGE: The types of some model class fields are extended from
    Optional[int] to Optional[int | FormatString], so downstream code
    that relies on those types can fail type checking.

No.

----


src/openjd/model/v2023_09/_model.py
```diff
@@ -285,7 +292,31 @@ class CancelationMethodNotifyThenTerminate(OpenJDModel_v2023_09):
...
+                # Try to parse as int, fail if not
+                try:
+                    return int(v)
+                except ValueError:
+                    raise ValueError(
+                        "notifyPeriodInSeconds as a format string requires the FEATURE_BUNDLE_1 extension."
+                    )
+            return validate_int_fmtstring_field(v, ge=1, context=context)
+        if isinstance(v, int):
+            if v < 1 or v > 600:
```
jericht, line 327: Is this 600 max value new?

jericht (dismissed):

test/openjd/model/test_parse.py
```diff
@@ -194,13 +194,15 @@ class MockExtensionName(str, Enum):
...
     SUPPORTED_NAME = "SUPPORTED_NAME"
+    FEATURE_BUNDLE_1 = "FEATURE_BUNDLE_1"
 
 
 class MockExtensionNameWithTwoNames(str, Enum):
     """A mock enum with only SUPPORTED_NAME for testing."""
 
     SUPPORTED_NAME = "SUPPORTED_NAME"
     ANOTHER_SUPPORTED_NAME = "ANOTHER_SUPPORTED_NAME"
+    FEATURE_BUNDLE_1 = "FEATURE_BUNDLE_1"
```
epmog, line 205: should these two enums be changed? Docstring indicates otherwise

src/openjd/model/v2023_09/_model.py
```diff
@@ -187,17 +201,19 @@ class ValueReferenceConstants(Enum):
...
+    # because the template name can contain format strings
 
     def __new__(cls, value: str, *, context: ModelParsingContextInterface = ModelParsingContext()):
         return super().__new__(cls, value, context=context)
 
 
 JobName = Annotated[
     str,
-    StringConstraints(min_length=1, max_length=128, strict=True, pattern=_standard_string_regex),
+    StringConstraints(min_length=1, max_length=512, strict=True, pattern=_standard_string_regex),
```
sakshie95, line 213: Have you kept this as a buffer cz of comment in the code above?

```
# Max length is validated after resolution in Job model, not here
# because the template name can contain format strings
```

src/openjd/model/v2023_09/_model.py
```diff
@@ -415,12 +493,91 @@ class EmbeddedFileText(OpenJDModel_v2023_09):
...
+            raise ValueError(f"name must be at most {max_len} characters long")
+        return v
+
+    @field_validator("filename")
+    @classmethod
+    def _validate_filename(cls, v: Optional[Filename], info: ValidationInfo) -> Optional[Filename]:
+        if v is None:
+            return v
+        context = cast(Optional[ModelParsingContext], info.context)
+        max_len = 256 if context and "FEATURE_BUNDLE_1" in context.extensions else 64
```
sakshie95, line 511: nit: maybe using constants for the constraints would be better here so its easy to update

src/openjd/model/v2023_09/_model.py
```diff
@@ -1052,6 +1209,15 @@ class Environment(OpenJDModel_v2023_09):
 
     _template_variable_scope = ResolutionScope.SESSION
 
+    @field_validator("name")
+    @classmethod
+    def _validate_environment_name(cls, v: str, info: ValidationInfo) -> str:
+        context = cast(Optional[ModelParsingContext], info.context)
+        max_len = 512 if context and "FEATURE_BUNDLE_1" in context.extensions else 64
```
sakshie95, line 1216: nit: Same as above, maybe using constants would be better here

src/openjd/model/v2023_09/_model.py
mwiebe: It's not, you can see it in specified and then applied here:

* https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#53-cancelationmethod
* https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/src/openjd/model/v2023_09/_model.py#L254

src/openjd/model/v2023_09/_model.py
mwiebe: The bounds that we put in the type annotations always apply when validating/parsing an input with pydantic, so the upper limit has to be the maximum of upper limits supported. Later, it is dynamically checked against the extension-specific limit.

src/openjd/model/v2023_09/_model.py
mwiebe: I think when you apply LLM code prompting, this convention that was true before may be reversed now - with the values directly here there are fewer indirections that are necessary to cross-validate against the written spec, for example.

src/openjd/model/v2023_09/_model.py
```diff
@@ -398,13 +473,16 @@ class EmbeddedFileText(OpenJDModel_v2023_09):
             will have its execute-permissions set.
             Default: False
         data (FormatString): The text data to write to the file.
+        endOfLine (Optional[EndOfLine]): Line ending style. Requires FEATURE_BUNDLE_1 extension.
+            Default: AUTO
```
epmog, line 477: do we need to describe AUTO? Is it based on operating system Windows vs. Posix? Or extension type? I don't see it listed in the RFC

src/openjd/model/v2023_09/_model.py
sakshie95: Ah well this is subjective I would rather teach the LLMs to follow the convention to keep the code clean :)) but non-blocking nonetheless, approved!

sakshie95 (approved):

jericht (approved):

test/openjd/model/test_parse.py
mwiebe: Good call, that was in there because of a different change that was fixed later.

src/openjd/model/v2023_09/_model.py
```diff
@@ -2785,10 +3161,13 @@ def _permitted_extension_names(
...
     @field_validator("parameterDefinitions")
     @classmethod
-    def _unique_parameter_names(
-        cls, v: Optional[JobParameterDefinitionList]
+    def _validate_parameter_definitions(
+        cls, v: Optional[JobParameterDefinitionList], info: ValidationInfo
     ) -> Optional[JobParameterDefinitionList]:
         if v is not None:
+            # EnvironmentTemplate always has max 50 parameters (no extension increases this)
+            if len(v) > 50:
```
leongdl, line 3169: Super nit - do we put constants like this somewhere ? Could be useful as an exported param?

src/openjd/model/v2023_09/_model.py
mwiebe: Yeah, it's worthwhile describing that here. It's specified in https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#61-embeddedfiletext

You're right that it's not in https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0004-enhanced-limits-and-capabilities.md. It was, and then we noted how clunky it was to include the block diffs in the RFC and shifted them out into the PR itself. But in doing so, we never added descriptions of those changes back into the RFC. I'm inclined to leave it as is and proceed with what we have in the spec, and watch for this phenomenon in future RFCs.

src/openjd/model/v2023_09/_model.py
```diff
@@ -2605,10 +2973,18 @@ def _unique_step_names(cls, v: StepTemplateList) -> StepTemplateList:
...
     @classmethod
-    def _unique_parameter_names(
-        cls, v: Optional[JobParameterDefinitionList]
+    def _validate_parameter_definitions(
+        cls, v: Optional[JobParameterDefinitionList], info: ValidationInfo
     ) -> Optional[JobParameterDefinitionList]:
         if v is not None:
+            # Validate max length based on extension
+            context = cast(Optional[ModelParsingContext], info.context)
+            max_len = 200 if context and "FEATURE_BUNDLE_1" in context.extensions else 50
```
leongdl, line 2982: The customer will be happy!

leongdl (approved):

src/openjd/model/v2023_09/_model.py
mwiebe: The source of truth for that is here: https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#12-environment-template

The conformance review prompt that I showed in the PR description did note this value and that it matched the spec in the report it produced. I also think the conformance suite in https://github.com/OpenJobDescription/openjd-specifications/pull/103 is a good place documenting it. These feel sufficient to me, so I'd rather not add another version of these constants in another separate place.

src/openjd/model/v2023_09/_model.py
mwiebe: I'll merge this now, and then fix up your two comments in a trivial follow-up

PR #263, opened 2026-02-11
fix: reject control characters in job names

crowecawcaw:
The spec bans control characters in job names, but this package allowed them.

Spec reference: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#111-jobname

Add a regex to reject them. 

Add unit tests. Also by running the conformance tests against the changes: https://github.com/OpenJobDescription/openjd-specifications/pull/103

n/a

No

No
----

### DIFF
--- a/src/openjd/model/v2023_09/_model.py
+++ b/src/openjd/model/v2023_09/_model.py
@@ -203,6 +203,8 @@ class JobTemplateName(FormatString):
     _min_length = 1
     # Max length is validated after resolution in Job model, not here
     # because the template name can contain format strings
+    # All unicode except the [Cc] (control characters) category
+    _regex = f"(?-m:^[^{_Cc_characters}]+\\Z)"
 
     def __new__(cls, value: str, *, context: ModelParsingContextInterface = ModelParsingContext()):
         return super().__new__(cls, value, context=context)
--- a/test/openjd/model/v2023_09/test_job_template.py
+++ b/test/openjd/model/v2023_09/test_job_template.py
@@ -197,6 +197,33 @@ class TestJobTemplate:
             ),
             pytest.param(
                 {
+                    "specificationVersion": "jobtemplate-2023-09",
+                    "name": "Job\x1fName",
+                    "steps": [STEP_TEMPLATE],
+                },
+                1,
+                id="name with control char 0x1F",
+            ),
+            pytest.param(
+                {
+                    "specificationVersion": "jobtemplate-2023-09",
+                    "name": "Job\x7fName",
+                    "steps": [STEP_TEMPLATE],
+                },
+                1,
+                id="name with control char 0x7F",
+            ),
+            pytest.param(
+                {
+                    "specificationVersion": "jobtemplate-2023-09",
+                    "name": "Job\x9fName",
+                    "steps": [STEP_TEMPLATE],
+                },
+                1,
+                id="name with control char 0x9F",
+            ),
+            pytest.param(
+                {
                     "name": "Foo",
                     "steps": [STEP_TEMPLATE],
                 },
--- a/test/openjd/model/v2023_09/test_job_template.py
+++ b/test/openjd/model/v2023_09/test_job_template.py
@@ -224,6 +224,24 @@ class TestJobTemplate:
             ),
             pytest.param(
                 {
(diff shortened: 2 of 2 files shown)


src/openjd/model/v2023_09/_model.py
```diff
@@ -203,6 +203,8 @@ class JobTemplateName(FormatString):
     _min_length = 1
     # Max length is validated after resolution in Job model, not here
     # because the template name can contain format strings
+    # All unicode except the [Cc] (control characters) category
+    _regex = f"(?-m:^[^{_Cc_characters}]+\\Z)"
```
mwiebe, line 207: What's the difference between this and `_standard_string_regex`?

crowecawcaw: The model uses two regex engines: one is Rust's that's embedded in Pydantic and the other is Python's. They have slightly different syntax. `_standard_string_regex` is formatted for Pydantic/Rust's engine with `\z` and this regex is formatted for Python with `\Z`. This regex in this class is eventually evaluated here: https://github.com/OpenJobDescription/openjd-model-for-python/blob/96651fa9eec05d0fa222f0e5bf8b394cd0b2ca34/src/openjd/model/_format_strings/_dyn_constrained_str.py#L55

Here's another example of Python regexes in use in the model:  
https://github.com/OpenJobDescription/openjd-model-for-python/blob/96651fa9eec05d0fa222f0e5bf8b394cd0b2ca34/src/openjd/model/v2023_09/_model.py#L250

Not sure there's a good reason for it. I'm guessing it is an outcome of upgrading from Pydantic v1 to v2.

joel-wong-aws: +1. From Kiro:
```
_standard_string_regex = rf"(?-m:^[^{_Cc_characters}]+\z)"

\\Z (uppercase Z): Matches at the end of the string OR just before a newline at the end

"hello\n" ✅ matches
"hello" ✅ matches

\z (lowercase z): Matches only at the absolute end of the string

"hello\n" ❌ does not match
"hello" ✅ matches
```
The lowercase `\z` seems to be the correct choice here.

leongdl:
I think we should add a fuzz test suite to OpenJD and have broader testing. Maybe a good backlog item for the OpenJD team.

src/openjd/model/v2023_09/_model.py
mwiebe: I remember there being something different between `\z` and `\Z` between pydantic v2 and Python regexes, because pydantic v2 switched to using the Rust regex crate.

crowecawcaw: The `\Z` vs `\z` difference is intentional due to two different regex engines being in play:

- `_standard_string_regex` (used by `JobName`, `StepName`, `EnvironmentName` via pydantic `StringConstraints`) uses lowercase `\z` — this is correct because pydantic v2 validates these patterns via `pydantic-core`, which uses the **Rust `regex` crate**. In Rust regex, `\z` is the absolute end-of-string anchor. [Ref](https://docs.rs/regex/latest/regex/#empty-matches)

- `JobTemplateName._regex` (used by `DynamicConstrainedStr._validate`) uses uppercase `\Z` — this is correct because validation runs through **Python's `re.match()`**. In Python's `re` module, `\Z` matches only at the absolute end of the string (it does NOT match before a trailing newline, unlike `$`). Python doesn't support lowercase `\z` in versions before 3.14 — it throws `re.PatternError: bad escape \z`. [Ref](https://docs.python.org/3/library/re.html)

So both are equivalent in behavior (absolute end-of-string), just spelled differently for their respective engines.

I've added additional tests to verify newlines are handled correctly.

crowecawcaw:
We actually have a new-ish fuzzer that uncovered a couple issues: https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/test/openjd/model/test_fuzz.py

We could definitely extend it for better coverage. In this particular case, I don't think it would have helped because the problem was not that a regex was faulty, it's that the library was missing a validation.

joel-wong-aws (approved):

mwiebe (approved):

PR #265, opened 2026-02-12
fix!: require attribute and amount names to be unique

crowecawcaw:
The amount and attribute names in host requirements are not required to be unique at the moment; duplicate names are allowed. Multiple requirements with the same name is non-sensical though and should be rejected.

Related OpenJD spec PR: https://github.com/OpenJobDescription/openjd-specifications/pull/110

While updating the test, I noticed existing test was incorrectly set up and testing the wrong data model. Since the test just looked for failures, the test passed but the validation failures were for the wrong reason. I made the test more specific to ensure it's testing what we think we're testing.

There are other tests that make similar mistakes and will also need to be updated. Future work.

Reject a template if a step's host requirements has duplicated attribute or amount names.

Closes a undefined behavior gap.

Unit tests

n/a

Technically yes, but practically no. Duplicated requirements do not have a clear behavior, and templates with duplicates are likely buggy. Marking the PR with `fix!` to indicate a minor version bump is recommended.

No

----

### DIFF
--- a/test/openjd/model/v2023_09/test_step_host_requirements.py
+++ b/test/openjd/model/v2023_09/test_step_host_requirements.py
@@ -457,24 +457,26 @@ class TestHostRequirementsTemplate:
         # no exception was raised.
 
     @pytest.mark.parametrize(
-        "data,error_count",
+        "data,error_count,error_loc",
         [
-            pytest.param({}, 1, id="missing amounts & attributes"),
-            pytest.param({"unknown": "value"}, 1, id="unknown field"),
-            pytest.param({"amounts": []}, 1, id="too few amounts"),
-            pytest.param({"attributes": []}, 1, id="too few attributes"),
+            pytest.param({}, 1, (), id="missing amounts & attributes"),
+            pytest.param({"unknown": "value"}, 1, ("unknown",), id="unknown field"),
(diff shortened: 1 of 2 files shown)


test/openjd/model/v2023_09/test_step_host_requirements.py
```diff
@@ -485,16 +487,66 @@ def test_parse_success(self, data: dict[str, Any]) -> None:
...
+                    ]
+                },
+                1,
+                ("amounts",),
+                id="duplicate amount names",
+            ),
+            pytest.param(
+                {
+                    "amounts": [
+                        {"name": "amount.mycap", "min": 1},
```
leongdl, line 507: Does the openjd spec allow this or specify properties w/ different cases are duplicates ?

leongdl: Trying to find it in the spec and it escapes me right now.

crowecawcaw: They are case insensitive! https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#3311-amountcapabilityname

leongdl (approved):

mwiebe (approved):

PR #268, opened 2026-02-18
fix: enforce maximum of 1024 items in range expressions

crowecawcaw:
Range expressions should only allow 1024 items, but this package doesn't enforce the limit.

https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#3411-inttaskparameterdefinition

Enforce the limit.

Library matches spec closer.

Unit tests

n/a

No

No
----

### DIFF
--- a/src/openjd/model/_range_expr.py
+++ b/src/openjd/model/_range_expr.py
@@ -150,6 +150,10 @@ class IntRangeExpr(Sized):
 
     def _validate(self) -> None:
         """raises: ValueError - if not valid"""
+        if len(self) > 1024:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


mwiebe:
I recall we didn't have the same low limit for them on purpose, as it's a much more efficient way to express 10,000 or 100,000 task jobs. A limit makes sense, maybe a million is better?

Reviewing the spec, I see the 1024 limit on `<IntRangeList>` but not on `<IntRangeExpr>`.

crowecawcaw:
I see. Closing this PR. Created a spec issue for clarification: https://github.com/OpenJobDescription/openjd-specifications/issues/114

PR #285, opened 2026-06-18
feat: Implement WRAP_ACTIONS with EXPR binding to RS

leongdl:
Fixes: *<insert link to GitHub issue here>*

The pure-Python (v0) model did not support the **EXPR** extension (RFCs 0005–0007) or the **WRAP_ACTIONS** extension (RFC 0008). As a result it diverged from the `openjd-rs` reference implementation and from the published specification:

- Templates using EXPR expressions, RFC 0007 typed job parameters, or RFC 0008 wrap-action hooks (`onWrapEnvEnter` / `onWrapTaskRun` / `onWrapEnvExit`) could not be parsed or validated by the Python model.
- Any expression/validation logic added on the Python side risked drifting from the authoritative Rust implementation.
- Downstream consumers (e.g. `openjd-sessions-for-python`) had no way to obtain the wrap-hook model or the typed EXPR symbol-table machinery they depend on.

- Wire the pure-Python model's format-string / expression handling to the Rust `openjd-expr` engine through the `openjd._openjd_rs` PyO3 bindings, instead of reimplementing the expression grammar in Python.
- Add focused bindings so the Rust logic is **reused, not duplicated**:
  - `build_symbol_table` — typed, dotted-key EXPR symbol tables (string→typed coercion lives next to the engine).
  - `job_parameter_type_expr_spec` — single-sourced OpenJD-type → EXPR type-spec mapping (including `LIST[...]` nesting), replacing a hand-maintained Python table.
  - `ParsedExpression.typecheck` — typed static validation without result extraction (removes brittle error-string sniffing).
- Implement v0 parity for **RFC 0007** typed parameters (typed defaults flowing as native values) and **RFC 0008** wrap-action rules, mirroring the `openjd-rs` validators:
  - the single-wrap-layer constraint,
  - all-or-nothing wrap-hook definition (all three hooks required together),
  - per-hook `WrappedAction.*` / `WrappedEnv.*` / `WrappedStep.*` variable scoping.
- Scope-check timeout format strings and add RANGE_EXPR typed-validation coverage.
- Resolve pre-existing `mypy` errors so the package type-checks cleanly.

The Python model now accepts and validates EXPR (RFC 0005–0007) and WRAP_ACTIONS (RFC 0008) templates consistently with the `openjd-rs` reference and the specification. Expression parsing/evaluation and the OpenJD-type → EXPR-type mapping are single-sourced in Rust, eliminating Python/Rust drift. This unblocks downstream consumers (notably `openjd-sessions-for-python`) that require the wrap-hook model and the EXPR bindings.

See [DEVELOPMENT.md](https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/DEVELOPMENT.md#testing) for information on running tests.

- Have you run the unit tests? **Yes.**

The full unit **and** conformance test suite passes: **5273 passed, 24 skipped, 6 xfailed** (the 6 `xfail`s are pre-existing, documented known gaps, not regressions). All conformance fixtures pass, including the WRAP_ACTIONS cases. Static checks are also clean: `ruff check`, `black --check`, and `mypy` (105 source files) all pass.

- Are relevant docstrings in the code base updated? **Yes.** New and changed modules/functions carry docstrings (e.g. the `_expr_support` Rust-bridge module, the new binding methods, and the wrap-rule validators reference the corresponding RFCs). No user-facing schema/doc changes were required beyond code-level documentation.

A breaking change is one that modifies a public contract in a way that is not backwards compatible. See the [Public Interfaces](https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/DEVELOPMENT.md#the-packages-public-interface) section of the DEVELOPMENT.md for more information on the public contracts.

**No.** The changes are additive and opt-in: EXPR and WRAP_ACTIONS behavior is enabled only when those extensions are declared on a template. Existing (non-EXPR) templates parse and validate exactly as before. Public type aliases were widened (never narrowed), so existing callers remain compatible.

- Does the change need to be threat modeled? **No new threat model is required for the model package.** User-supplied template expressions are evaluated by the Rust `openjd-expr` engine (bounded memory, no arbitrary code execution) and only when the EXPR extension is declared. This change does not create or modify any files/directories that must be readable only by the process owner.
- If so, then please label this pull request with the "security" label. We'll work with you to analyze the threats.

----


src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+"""Support for the EXPR extension (RFCs 0005/0006/0007) in the pure-Python
+(v0) model.
+
+The pure-Python model does not implement the EXPR expression grammar. When a
+template declares the ``EXPR`` extension, format-string expressions are parsed
+and evaluated by the Rust ``openjd-expr`` engine through the ``openjd._openjd_rs``
+bindings. This module is the thin bridge between the two.
+
+Design reference:
+``SuperDaveDocs/.../delivery-estimate-v3/rs/python-model-with-rust-expr``.
```
leongdl: Clean up - this is not needed in comment.

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+    RangeExprError as RustRangeExprError,
+)
+
+EXPR_EXTENSION = "EXPR"
+WRAP_ACTIONS_EXTENSION = "WRAP_ACTIONS"
+
+# Errors raised by the Rust expr engine. All subclass ValueError but are
+# distinct classes from the model's own ExpressionError, so they must be
+# caught explicitly at the binding boundary and re-raised as model errors.
+_RUST_EXPR_ERRORS = (
```
leongdl, line 50: Do the errors here match the PyO3 errors surfaced up by openjd model _v1 namespace - they should match 1:1

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+      int/float/bool/str/list at the boundary.
+    """
+    if type_name is None:
+        return value
+
+    expr_type = _OPENJD_TYPE_TO_EXPR_TYPE.get(type_name.upper())
+    if expr_type is None:
+        # Unknown/aggregate type (e.g. LIST[*], RANGE_EXPR) — let the engine
+        # infer from the native value for now. Typed list/range coercion is
+        # follow-up work (RFC 0007 parameter types).
```
leongdl: Why? I thought it was implemented.

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+        return value
+    if expr_type == "path":
+        return ExprValue(str(value), type=ExprType("path"), path_format=path_format)
```
leongdl: Do we have equivalent test cases compared to openjd-rs for the path part.

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+    EXPR engine can consume.
+
+    The v0 SymbolTable uses flat dotted keys (``"Param.Frame"``) and direct
+    lookup; the Rust ``dict_to_symtab`` builds subtables from nested dicts. We
+    split each dotted key on ``.`` (parameter names cannot contain ``.``) and
+    nest accordingly, re-typing each leaf from ``types`` when available.
+    """
+    types = types or {}
+    nested: dict[str, Any] = {}
+    for dotted_name in symtab.symbols:
```
leongdl: Improvement - can we implement this in rust.

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def parse_expr_or_raise(expr: str) -> Any:
+    """Parse an EXPR expression string with the Rust engine, re-raising parse
+    failures as the model's ``ExpressionError``.
+    """
+    try:
+        return parse_expression(expr.strip())
+    except _RUST_EXPR_ERRORS as exc:
+        raise _ModelExpressionError(str(exc))
```
leongdl, line 94: Is this exception 1:1 the same as the Rust version?

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+    """Map an OpenJD parameter type name (e.g. ``PATH``, ``LIST[INT]``) to the
+    EXPR engine's spec-form type string (``path``, ``list[int]``), or ``None``
+    if there is no confident mapping (so callers fall back to name-only
+    validation rather than risk a wrong type).
+    """
+    t = openjd_type.strip().upper()
+    scalar = _OPENJD_TYPE_TO_EXPR_TYPE.get(t)
+    if scalar is not None:
+        return scalar
+    if t.startswith("LIST[") and t.endswith("]"):
```
leongdl: This seems fragile, can we add a constant?

src/openjd/model/_format_strings/_expr_support.py
```diff
@@ -0,0 +1,237 @@
...
+        inner = expr_type_for_openjd_type(t[len("LIST[") : -1])
+        if inner is not None:
+            return f"list[{inner}]"
+    return None
+
+
+def longest_defined_prefix(name: str, defined: set) -> Optional[str]:
```
leongdl, line 145: Make sure any function not in a _ prefixed file name has proper function prefixes with `_` so we do not leak any public functions.

src/openjd/model/_format_strings/_nodes.py
```diff
@@ -79,10 +87,122 @@ def validate_symbol_refs(self, *, symbols: set[str]) -> None:
...
+    def evaluate(self, *, symtab: SymbolTable, path_format: Any = None) -> Any:
         if self.name not in symtab:
             raise ValueError(f"{self.name} has no value")
         return symtab[self.name]
 
     def __repr__(self):
         return f"FullName({self.name})"
+
+
+class ExprNode(Node):
```
leongdl, line 119: Is the coding style to embed all the nodes in one file? Or should this be in another dedicated file?

leongdl, line 119: This node only creates if EXPR and WRAP_ACTIONS is defined in the template ? Check this.

src/openjd/model/_format_strings/_nodes.py
```diff
@@ -79,10 +87,122 @@ def validate_symbol_refs(self, *, symbols: set[str]) -> None:
...
+
+        # Typed validation: if every accessed symbol resolves to a defined
+        # symbol prefix whose EXPR type is known (e.g. job parameters), validate
+        # the whole expression against those types. This catches type mismatches
+        # and resolves method/property access (e.g. Param.File.name on a PATH).
+        # Only attempted when ALL accessed prefixes are typed-known, so an
+        # unknown-typed symbol (e.g. a `let` name) safely falls back to the
+        # name-only check below rather than risking a wrong-type rejection.
+        if symbol_types and accessed:
+            from ._expr_support import (
```
leongdl, line 168: Move these imports to __init__ so its in 1 place of failure.

src/openjd/model/_format_strings/_nodes.py
```diff
@@ -79,10 +87,122 @@ def validate_symbol_refs(self, *, symbols: set[str]) -> None:
...
+
+        # accessed_symbols are the free variable references in full dotted
+        # spelling (e.g. "Param.X"); local let-bound names are excluded by the
+        # engine. Compare against the set of symbols visible at this location.
+        missing = accessed - symbols
+        if missing:
+            from ._edit_distance import closest
+
+            name = sorted(missing)[0]
+            msg = f"Variable {name} does not exist at this location."
```
leongdl: Are these error messages 1:1 the same as rust? make sure it is the same for seamless migration.

src/openjd/model/_format_strings/_nodes.py
```diff
@@ -79,10 +87,122 @@ def validate_symbol_refs(self, *, symbols: set[str]) -> None:
...
+        shadowed = set(self._parsed.local_bindings) & symbols
+        if shadowed:
+            raise ValueError(
+                f"Variable {sorted(shadowed)[0]!r} bound by a comprehension shadows "
+                "a variable that already exists at this location."
+            )
+
+    def evaluate(self, *, symtab: SymbolTable, path_format: Any = None) -> Any:
+        from ._expr_support import (
+            ExprProfile,
```
leongdl, line 214: move import to init.

src/openjd/model/_format_strings/_parser.py
```diff
@@ -24,6 +30,9 @@ def parse_format_string_expr(expr: str, *, context: ModelParsingContextInterface
     Returns:
         Node: Root of the expression tree.
     """
+    if EXPR_EXTENSION in context.extensions:
+        # Raises: model ExpressionError on parse failure.
+        return ExprNode(expr)
```
leongdl, line 40: That was an easy integration!

src/openjd/model/_internal/_variable_reference_validation.py
```diff
@@ -408,14 +477,111 @@ def _check_format_string(
...
+        name, sep, rhs = binding.partition("=")
+        name = name.strip()
+        rhs = rhs.strip()
+        if not sep or not name or not rhs:
+            # Malformed binding; the `let` field validator reports the details.
+            continue
+        binding_loc = (*loc, "let", index)
+        if name in enclosing:
+            errors.append(
+                InitErrorDetails(
```
leongdl, line 563: Do these errors match openjd-rs 1:1

src/openjd/model/v2023_09/_model.py
```diff
@@ -348,6 +359,9 @@ class CancelationMethodTerminate(OpenJDModel_v2023_09):
 
 ArgListType = Annotated[list[ArgString], Field(min_length=1)]
 
+# WRAP_ACTIONS (RFC 0008) wrap-hook field names on EnvironmentActions.
+_WRAP_ACTION_FIELDS = ("onWrapEnvEnter", "onWrapTaskRun", "onWrapEnvExit")
```
leongdl, line 363: Are constants declared inline across the model generally?

src/openjd/model/v2023_09/_model.py
```diff
@@ -414,20 +428,63 @@ class EnvironmentActions(OpenJDModel_v2023_09):
...
 
+        context = cast(Optional[ModelParsingContext], info.context) if info else None
+        extensions = context.extensions if context else set()
+
+        wrap_values = {name: values.get(name) for name in _WRAP_ACTION_FIELDS}
+        any_wrap = any(v is not None for v in wrap_values.values())
+
+        if any_wrap:
+            if "WRAP_ACTIONS" not in extensions:
+                raise ValueError(
```
leongdl: Make sure the error message is the same as openjd-rs

src/openjd/model/v2023_09/_model.py
```diff
@@ -414,20 +428,63 @@ class EnvironmentActions(OpenJDModel_v2023_09):
...
+                    "The onWrapEnvEnter, onWrapTaskRun, and onWrapEnvExit actions "
+                    "require the WRAP_ACTIONS extension."
+                )
+            if "EXPR" not in extensions:
+                raise ValueError("The WRAP_ACTIONS extension requires the EXPR extension.")
```
leongdl: Make sure all error messages match openjd-rs

src/openjd/model/v2023_09/_model.py
```diff
@@ -550,6 +607,59 @@ class ScriptInterpreter(str, Enum):
     NODE = "node"
 
 
+LET_MAX_BINDINGS = 50
```
leongdl, line 720: Does this match the openjd spec?

rust-bindings/src/expr/mod.rs
```diff
@@ -26,5 +26,5 @@ pub(crate) use path_mapping::PyPathMappingRule;
 pub(crate) use profile::{PyExprExtension, PyExprProfile, PyExprRevision, PyHostContext};
 pub(crate) use range_expr::{PyIntRange, PyRangeExpr};
 pub(crate) use symbol_table::{
-    _reconstruct_serialized_symtab, PySerializedSymbolTable, PySymbolTable,
+    _reconstruct_serialized_symtab, build_symbol_table, PySerializedSymbolTable, PySymbolTable,
```
leongdl, line 29: # PR #285 — EXPR / WRAP_ACTIONS: key call stacks

**Repo:** OpenJobDescription/openjd-model-for-python · branch `python-rs`
**What the PR does:** teaches the pure-Python (v0) model to handle the RFC
0005–0008 expression extensions by delegating `{{ }}` parsing, type-checking,
and evaluation to the Rust `openjd-expr` engine over PyO3 bindings.

## TL;DR

Before this PR the pure-Python model only understood the legacy
`Name.Dot.Name` interpolation grammar. This PR makes it a thin front-end over
the Rust expression engine whenever a template declares the `EXPR` extension,
and adds the parameter types and wrap-action rules that ride on top of it.

- **EXPR expressions** — when `EXPR` is declared, `{{ }}` parsing/eval routes to
  a new Rust-backed `ExprNode` (arithmetic, functions, lists, paths, `let`,
  comprehensions) instead of the legacy name parser. Non-EXPR templates are
  byte-for-byte unchanged, and the Rust surface is imported lazily so the
  non-EXPR path never loads it.
- **Type-aware validation** — job-parameter types are mapped to EXPR types and
  fed to a new `ParsedExpression.typecheck()` that catches real type errors at
  decode time without false-rejecting runtime-dependent expressions (no more
  error-string sniffing).
- **8 new job-parameter types** (RFC 0007) — `BOOL`, `RANGE_EXPR`, and the
  `LIST[*]` family — with case-insensitive type names, item/length constraints,
  and an EXPR-extension gate, all sharing one base class.
- **Native values end-to-end** — `ParameterValue.value`/`JobParameter.value`
  widen from `str` to `Any` so lists/bools survive `create_job` intact; a new
  `SymbolTable.expr_types` field carries the type tags so the Rust builder
  coerces correctly (and is preserved across copy/union).
- **Spec-correct string coercion** — format-string output now uses the engine's
  RFC 0005 coercion via `evaluate_to_str()` (`{{ true }}` → `"true"`, null →
  `""`, quoted list items) instead of Python's `str()`.
- **WRAP_ACTIONS (RFC 0008)** — `onWrapEnvEnter/onWrapTaskRun/onWrapEnvExit` on
  `EnvironmentActions`, gated on `WRAP_ACTIONS`+`EXPR`, all-or-nothing, with
  per-hook `Wrapped*` variable scoping and a job-wide single-wrap-layer rule.
- **New Rust bindings** — `build_symbol_table`, `ParsedExpression.typecheck`,
  `job_parameter_type_expr_spec` (single-sources the OpenJD→EXPR type mapping).

**Risk shape:** almost all behavior change is gated behind the `EXPR` /
`WRAP_ACTIONS` extension declarations, so templates that don't opt in are
unaffected. The sharp edges are at the Python↔Rust boundary (the three flows
that cross it) and the `str → Any` value widening.

---

Two flows form the spine of the change: **(1)** decode-time validation of an
expression's symbol references, and **(2)** runtime resolution of an expression
into the string substituted into a format string. `──BOUNDARY──▶` marks a call
crossing from Python into compiled Rust; `◀──` marks the return.

---

## Flow 1 — Decode-time: validate an EXPR `{{ }}` expression's symbols

Entry: `decode_job_template(...)` → pydantic validation → the variable-reference
prevalidator.

```
Flow: prevalidate_model_template_variable_references(values, context)
  1. _variable_reference_validation.py:137   prevalidate_model_template_variable_references
  2.   :166                                   expr_enabled = "EXPR" in declared & in context.extensions
  3.   :173                                   context._symbol_types = _build_job_parameter_symbol_types(values)
  4.     └─ :433                              _build_job_parameter_symbol_types
  5.        └─ _expr_support.py               expr_type_for_openjd_type("LIST[INT]")
  6.           └──BOUNDARY──▶                 _openjd_rs job_parameter_type_expr_spec   (model/types.rs)
  7.                                          uppercase-parse + OpenJD→EXPR map   ◀── "list[int]" | None
  8.   :460                                   _check_format_string(value, scope, symbols, context)
  9.     :484                                 expr.validate_symbol_refs(symbols, symbol_types=ctx._symbol_types)
 10.        └─ _expression.py                 InterpolationExpression → self._expresion_tree.validate_symbol_refs
 11.           └─ _nodes.py:149              ExprNode.validate_symbol_refs(symbols, symbol_types)
 12.              :150                        accessed = self._parsed.accessed_symbols   (free dotted refs, from Rust)
 13.              :167                        for name: longest_defined_prefix(name, symbols) → prefix_types
 14.              :173    ┌── fully typed ──  validate_typed_expression(self._parsed, typed_symbols=prefix_types)
 15.              │       └─ _expr_support.py  values = {dotted: ExprValue.unresolved(ExprType(t))}
 16.              │          └──BOUNDARY──▶    _openjd_rs ParsedExpression.typecheck   (parsed_expression.rs)
 17.              │                            eval vs unresolved placeholders, DISCARD result,
 18.              │                            raise ONLY on a genuine type error   ◀──
 19.              └── else (some prefix untyped) ─ name-only fallback:
 20.                    :182                   missing = accessed - symbols
 21.                    :184                   if missing: raise _missing_symbol_error(name, symbols)
```

**Key insight:** step 16's `typecheck()` discards the result, so a "well-typed
but unresolved at runtime" expression passes cleanly — the model no longer has
to sniff Rust error-message strings to tell a real type error from a
runtime-dependent one. (The `let`-binding path reaches the same
`ExprNode.validate_symbol_refs` from `_variable_reference_validation.py:606`,
building one `ExprNode` per binding.)

---

## Flow 2 — Runtime: resolve a `{{ }}` expression to its substituted string

Entry: a caller invokes `FormatString.resolve(symtab=...)` during job
instantiation / argument resolution.

```
Flow: FormatString.resolve(symtab, path_format)  —  {{ }} → substituted string
  1. _format_string.py:79             FormatString.resolve(symtab, path_format)
  2.   :115                           element.expression.evaluate_to_str(symtab, path_format)
  3.     └─ _expression.py            InterpolationExpression → self._expresion_tree.evaluate_to_str
  4.        └─ _nodes.py:234          ExprNode.evaluate_to_str(symtab, path_format)
  5.           :241 / :208            value = self._evaluate_raw(symtab, path_format)
  6.              └─ :221             values = symtab_to_expr_values(symtab, types=symtab.expr_types or None)
  7.                 └─ _expr_support.py   map OpenJD types → EXPR specs (Flow 1, steps 5–7)
  8.                    └──BOUNDARY──▶ _openjd_rs build_symbol_table(flat, expr_types, path_format)
  9.                                   coerce "10":INT → int, nest "Param.Frame"   ◀── typed SymbolTable
 10.              :225 └──BOUNDARY──▶  _openjd_rs ParsedExpression.evaluate(values, profile, path_format)  ◀── ExprValue
 11.           :242                    if value.is_null: return ""        # null → empty string (RFC 0005)
 12.           :244                    return str(value)                  # ExprValue.__str__ = engine spec coercion
 13.   :127                           resolved_list.append(resolved_value)   # NOTE: str() wrapper removed here
```

**Key insight:** step 12's `str(ExprValue)` is the fix for the
`{{ true }}` → `"True"` Python-repr bug — it uses the engine's RFC 0005 coercion
(`true`/`false`, double-quoted list items, preserved `Decimal` trailing zeros),
not Python's `str()`. Step 9 is where the `expr_types` metadata earns its keep:
if that map is ever dropped on a `SymbolTable` copy/union, coercion silently
reverts to inference and produces a wrong-typed value — which is exactly why the
PR promotes `expr_types` to a real field that is copied in `__init__` and
`union`. `ExprNode.evaluate` (`_nodes.py:231`) instead calls `.item()` to hand
back the native Python value for callers that want the typed result.

---

## Flow 3 — Decode-time: a new typed job parameter (BOOL / LIST[*] / RANGE_EXPR)

The PR adds eight EXPR-extension job-parameter types. They share one base class
and a normalize-then-discriminate decode path. Entry: `decode_job_template(...)`
on a template whose `parameterDefinitions` contains, say, `{"type": "list[int]"}`.

```
Flow: JobTemplate.parameterDefinitions validation
  1. _model.py:3809  @field_validator("parameterDefinitions", mode="before")
  2.    └─ :3427      _normalize_parameter_type_case(value, info)
  3.       :3434      if "EXPR" in context.extensions:  item["type"] = item["type"].upper()
  4.                  # "list[int]" → "LIST[INT]"  (RFC 0007 case-insensitive type names)
  5.    │  (pydantic now resolves the discriminated union on the upper-cased "type")
  6.    └─ Literal[JobParameterType.LIST_INT] ─▶ JobListIntParameterDefinition
  7.       └─ class _JobListParameterDefinitionBase (_model.py:3477)
  8.          :3487   @field_validator("type")  _validate_type_gate → _expr_param_gate
  9.                  if context & "EXPR" not in extensions: raise "requires the EXPR extension"
 10.          :3519   @model_validator(after) _validate_default
 11.             └─ :3512  _check_list_value(self.default)
 12.                       _check_list_length(name, v, minLength, maxLength)
 13.                       for item: self._check_item(item)   # subclass hook
 14.                         └─ JobListIntParameterDefinition._check_item → _check_int_item
 15.                            └─ _check_numeric_item_bounds(...)   # shared int/float bounds (deduped this PR)
```

**Key insight:** every concrete list type is just a `type` literal + an `item`
constraint + a one-line `_check_item` override on the shared base — the length
check, EXPR gate, create-job metadata, and template-variable defs all live once
on `_JobListParameterDefinitionBase`. `RANGE_EXPR` (`_model.py:3644`) is the odd
one out: it validates its default string through the existing `IntRangeExpr`
grammar rather than a list check. Two-stage decode (normalize *before*,
discriminate *after*) is why a lowercase `list[int]` resolves to the right
class instead of failing the union.

---

## Flow 4 — Instantiation: create_job carries native values for the new types

Legacy scalars (STRING/INT/FLOAT/PATH) are stringified through preprocessing;
the new EXPR types must keep their native Python value (a real `list`/`bool`) so
the typed symbol-table builder (Flow 2, step 8) can coerce them. Entry:
`create_job(job_template, job_parameter_values)`.

```
Flow: create_job(template, values)
  1. _create_job.py:293   create_job
  2.   :64                _collect_defaults_2023_09
  3.     :79              is_legacy_scalar = param.type.name in _LEGACY_SCALAR_TYPE_NAMES
  4.     :82  ┌─ default ─ if not is_legacy_scalar:  ParameterValue(value=param.default)   # native, no str()
  5.     :123 └─ provided ─ if not is_legacy_scalar:  ParameterValue(value=value)          # native, no str()
  6.                       (legacy branch still does str(value) + PATH-relative join below)
  7.   :139                _check_2023_09(defs, values)
  8.     :155              check_constraints = getattr(param, "_check_constraints", None)
  9.                       if check_constraints: check_constraints(param_value.value)      # list/range checked here
 10.   :349                expr_types = {}; for each non-legacy param:
 11.     :357              expr_types["Param.<n>"] = expr_types["RawParam.<n>"] = param.type.value
 12.     :359              symtab.expr_types.update(expr_types)   # ← feeds Flow 2 step 8 coercion
```

**Key insight:** this is the producer side of the `expr_types` contract that
Flow 2 consumes. The widened `ParameterValue.value: Any` (was `str`) is what lets
a `[1, 2, 3]` survive instantiation intact; the review confirmed no downstream
consumer string-ops that value, and the type tag travels alongside it via
`symtab.expr_types` so the Rust builder coerces correctly.

### Merge wrinkle — `_merge_job_parameter.py`

When the same parameter is defined in multiple templates, only the four legacy
scalar types take the cross-merge path (`merge_job_parameter_definitions_for_one`
:191, gated on `_LEGACY_CONSTRAINT_TYPES`). EXPR types skip it and return the
last-defined definition via `model_copy` (:241) — which *bypasses validators*, so
the PR re-runs `getattr(base, "_check_constraints", None)` (:235) on the carried
default to avoid letting a constraint-violating default slip through.

---

## Flow 5 — Decode-time: WRAP_ACTIONS structural validation (RFC 0008)

Two independent checks fire when `WRAP_ACTIONS` is declared: per-`EnvironmentActions`
shape/scope, and a job-wide single-wrap-layer rule.

```
Flow: EnvironmentActions + JobTemplate WRAP_ACTIONS validation
  A. _model.py:516   EnvironmentActions._requires_oneof  (mode=before)
       any_wrap = any of (onWrapEnvEnter/onWrapTaskRun/onWrapEnvExit) set
       if any_wrap & context present:
         require "WRAP_ACTIONS" in extensions; require "EXPR" (hard prereq)
         require ALL THREE wrap hooks (all-or-nothing)
       # context-gated: skipped at create_job re-validation (no context)
  B. _model.py:565   EnvironmentActions._validate_wrapped_variable_scope (mode=after)
       for field in (onEnter, onExit, *wrap hooks):
         referenced = _action_referenced_namespaces(action)   # :398, scans command/args/timeout FormatStrings
            └─ expr.accessed_symbols → namespace = symbol.split(".")[0]   # WrappedAction/WrappedEnv/WrappedStep
         reject namespaces not allowed in that hook (WrappedStep only in onWrapTaskRun, etc.)
  C. _model.py:3941  JobTemplate._validate_single_wrap_layer (mode=after)
       job_env_wraps = count of jobEnvironments with a wrap hook (_env_defines_wrap_hook :373)
       for each step: if job_env_wraps + step's stepEnvironment wraps > 1: reject
       # "at most one wrap layer reachable per session stack"
```

**Key insight:** the three checks live at different altitudes on purpose —
extension/all-or-nothing gating per `EnvironmentActions` (before), per-hook
variable scoping per `EnvironmentActions` (after), and the cross-cutting
single-layer count at `JobTemplate` (after, where it can see jobEnvironments +
all steps). The scope check (B) deliberately inspects the `timeout` FormatString
too, not just command/args, because `timeout` is a format string under
FEATURE_BUNDLE_1.

---

## Reference — type changes that ripple through the flows

| Change | File | Why it matters |
|---|---|---|
| `ParameterValue.value: str → Any` | `_types.py` | lets BOOL/LIST values ride through preprocessing natively (Flow 4); a stringified list can't be re-parsed into a typed list |
| `JobParameter.value: str → Any` | `v2023_09/_model.py` | same, on the instantiated-job side |
| `SymbolTable.expr_types: dict[str,str]` (new field) | `_symbol_table.py` | the type-tag map; copied in `__init__`/`union` so coercion (Flow 2 step 8) never silently reverts to inference |
| `ExpressionInfo.resolved_value: Optional[Union[Real,str]] → Optional[str]` | `_format_string.py` | `resolve()` now appends the value directly (str() wrapper dropped); `evaluate_to_str` guarantees a string |
| `Node.allows_nonscalar` + `evaluate_to_str()` (new) | `_nodes.py` | lets EXPR nodes return list/bool/null while the legacy `FullNameNode` stays str/Real-only |
| `validate_symbol_refs(..., symbol_types=...)` (new kwarg) | `_nodes.py` / `_expression.py` | threads the type map into the typed-validation fork (Flow 1 step 11) |
| `DefinesTemplateVariables.expr_inject` (new) | `_types.py` | EXPR-only symbol injection (e.g. `Step.Name`) applied by the prevalidator, which has the parsing context |
| New Rust bindings: `build_symbol_table`, `ParsedExpression.typecheck`, `job_parameter_type_expr_spec` | `rust-bindings/` | the three boundary calls the flows above cross into |

src/openjd/model/_internal/_variable_reference_validation.py
seant-aws: Do we still need to further make sure rust/python errors conform ?
This can be a later thing, just curious about your own comment. Seems like there's some gaps ?

seant-aws:
I think in general things look good. I pulled the changes and tested on my own, nothing significant that I think needs changing, one comment about existing comment

seant-aws (approved):

seant-aws:
also double checking, would there be a good e2e way to test sessions model together ?

jericht (approved):
Looks good overall, we can iterate on any small issues in future PRs

PR #296, opened 2026-06-29
chore: rename YamlLoader to SafeLoader

epmog:

Scanner flagging yaml.load instead of yaml.safe_load.

safe_load does what we do, but we also allow the c-extension if it exists: https://github.com/yaml/pyyaml/blob/34a9bf82357f4952d8f194a5a31f1c39743652d0/lib/yaml/__init__.py#L125

Probably won't stop it, but changed the import aliases to reflect we're using SafeLoader regardless of c-extension.

Scanner reports reduced.

```
hatch run fmt
hatch run lint
```

N/A
N/a

N/A

----


godobyte (approved):

godobyte:
Curious what's the context for this change, which scanner you're referring to?

seant-aws (approved):

PR #307, opened 2026-07-08
chore(deps): bump pytest, black, mypy, hypothesis with Py39 compat

jericht:
Supersedes #300, #301, #302, #303.

### Reformatting changes only

This PR only contains reformatting changes due to the `black` upgrade

Some dependabot PRs are failing due to py39 not being supported by our deps. We need to update the deps still and pin py39 versions to the last supported ones

Update dev dependencies to their latest versions while maintaining Python 3.9 support via environment markers:

- pytest: 8.4.* -> 9.1.* (Py39 stays on 8.4.*)
- black: 25.* -> 26.* (Py39 stays on 25.*)
- mypy: 1.19.* -> 2.2.* (Py39 stays on 1.19.*)
- hypothesis: ~=6.0 -> ~=6.156 (Py39 gets ~=6.141)

All four packages dropped Python 3.9 support in their latest major releases. This uses the same conditional pattern as openjd-cli.

**Also had to reformat the files due to new black version**

Updated dependencies

CI passes

No

No

No

----

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,10 @@ updates:
       day: "monday"
     commit-message: 
       prefix: "chore(deps):"
+    ignore:
+      # black 26+ drops Python 3.9 compat and produces different formatting
+      - dependency-name: "black"
+        update-types: ["version-update:semver-major"]
   - package-ecosystem: "github-actions"
     directory: "/" # Location of package manifests
     schedule:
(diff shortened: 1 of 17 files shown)


seant-aws (dismissed):
💾

leongdl (dismissed):

wyongzhi (dismissed):

leongdl (approved):

seant-aws (approved):

PR #312, opened 2026-07-15
fix: deliver mid-action cancels through SessionCancelHandle

seant-aws:
> **Reworked 2026-07-24.** The original PR pre-armed a parent `CancellationToken` and cancel-request channel via `set_cancel_parent_token` / `set_cancel_request_channel` — API proposed in openjd-rs#258's original form, which was superseded by `Session::cancel_handle()` (openjd-rs#260, released in openjd-sessions 0.4.0; #258 was distilled to the `terminate_delay` capping and merged separately). This rework replaces the fallback machinery with the handle: smaller diff, same fix.

## What was the problem?

`Session.cancel_action()` failed with "session is busy" whenever an action was actually running: the Rust `Session::cancel_action` needs `&mut Session`, but a background action thread owns the `Session` for exactly that window — the primary window in which callers want to cancel. Mid-action cancels (service-initiated cancels and `time_limit` expiries in Deadline Cloud's worker agent) never reached the job process.

## What is the fix?

Obtain a `SessionCancelHandle` at session construction and route the taken-session case through `handle.cancel(time_limit, mark_action_failed)`. The handle delivers the cancel to the in-flight action without touching the `Session`, following the action's declared cancelation method on both the direct and cross-user helper paths (openjd-sessions 0.4.0 also carries the `min(time_limit, terminate_delay)` grace capping from openjd-rs#258).

One semantic note: the handle cannot perform the `Running → Canceling` session-state transition (it does not own the `Session`), so the binding updates its state snapshot to `Canceling` directly — Python-side pollers keep observing the cancel-in-progress phase until the terminal state lands, preserving the contract that openjd-sessions-for-python#331 (merged) relies on.

## How was this change tested?

* `cargo check` / `clippy --all-targets` / `fmt` clean; `cargo test` (build/link gate) passes against the published crates.io `openjd-sessions 0.4.0` — no patch overrides.
* Full Python suite: 5443 passed; the 4 failures in `test/openjd/expr/test_operation_limit.py` reproduce identically on pristine `mainline` in the same environment (pre-existing, unrelated).
* End-to-end validation with the Deadline Cloud worker agent's differential Python-vs-Rust cancel test suite is the next step and will be reported on this PR.

### DIFF
--- a/rust-bindings/src/sessions/session.rs
+++ b/rust-bindings/src/sessions/session.rs
@@ -265,6 +265,11 @@ pub(crate) struct PySession {
     session: Arc<Mutex<Option<Session>>>,
     /// Snapshot updated after each state change, readable without blocking on the session.
     snapshot: Arc<Mutex<StateSnapshot>>,
+    /// Thread-safe cancel handle, obtained at construction. Delivers a cancel
+    /// to whichever action is running even while the `Session` value is owned
+    /// by a background action thread — the case where `Session::cancel_action`
+    /// is unreachable (it needs `&mut Session`).
+    cancel_handle: openjd_sessions::session::SessionCancelHandle,
 }
 
 impl PySession {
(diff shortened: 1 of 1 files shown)


leongdl (approved):

godobyte (approved):
Does it make sense to add unit or integ tests as a follow-up?

PR #332, opened 2026-08-18
feat: produce and transport resolved symbol tables from Python

leongdl:
Two commits, both in service of one use case: letting a **scheduler** produce a
resolved symbol table and hand it to a **worker on another host** — the flow
`SerializedSymbolTable`'s own docstring describes as "a scheduler serializes the
symbol table to JSON and sends it to a worker that may be on a different OS".

Today neither half of that is reachable from Python: there is no supported way to
get the transport text out, and `create_job` discards the tables it builds.

| Commit | Adds |
|---|---|
| `feat: add SerializedSymbolTable JSON accessors` | `to_json_str()` / `from_json_str()` |
| `feat: return resolved symbol tables from create_job` | `create_job_with_symbol_tables()` → `JobWithSymbolTables` |

Happy to split these into two PRs if you'd prefer to review them separately.

---

## 1. JSON accessors

`SerializedSymbolTable` exposes `from_symtab`, `to_symtab` and `__reduce__`. To
move the transport text across a boundary today, a caller has to reach into the
pickle protocol:

```python
_helper, (json_text,) = serialized.__reduce__()   # not a contract
```

or call the private module-level pickle helper to get back in:

```python
from openjd._openjd_rs import _reconstruct_serialized_symtab   # underscore
```

The Rust side already has the capability (`from_json_str`, and `serde_json` via the
type's `Serialize` impl). This exposes it.

**Design notes**

- **`from_json_str` validates syntax, not contents.** Malformed JSON raises
  `ValueError`; a well-formed document that is not a valid table is accepted and
  rejected later by `to_symtab`. That matches the existing split — `to_symtab` is
  where entry validation lives — and keeps `from_json_str` cheap. Tests pin both
  halves so the behaviour is deliberate.
- **`__reduce__` is unchanged**, still routing through
  `_reconstruct_serialized_symtab`, so existing pickles keep working. Happy to
  re-point it and drop the private helper if you prefer, but that changes what old
  pickled bytes reference.
- **Output is stable.** `SymbolTable`'s `Serialize` impl emits entries in canonical
  path order, which the type already relies on for `Eq`/`Hash`, so the text is safe
  to store and compare.

## 2. Returning the resolved symbol tables

`create_job` builds a job-scope symbol table, and `instantiate_model` extends it
per step with `Step.Name` and the step's template-scope `let` bindings
(`StepTemplate._extend_step_symtab`). Both are then discarded. openjd-rs keeps the
equivalent table on each `Step` as `resolved_symtab`; the Python model has no
equivalent, so a Python scheduler cannot produce what a Rust session consumes.

```python
result = create_job_with_symbol_tables(
    job_template=job_template, job_parameter_values=values
)
result.job                                  # same Job create_job returns
result.job_symbol_table                     # Param.*, RawParam.*, Job.Name
result.step_symbol_tables["render"]         # + Step.Name, + template-scope let
```

**Design notes**

- **Returned, not attached to the model.** Adding a field to the pydantic `Step`
  would change the model's serialized form, and `Step.dict()` output is persisted
  as a worker contract by at least one consumer. A separate return value avoids
  that entirely.
- **`create_job` is untouched behaviourally.** Its body moved to
  `_create_job_and_symbol_table`, which both entry points call. A test asserts the
  two produce equal jobs.
- **The per-step tables come from the model's own hook.** Rather than
  reimplementing the step scope, this re-invokes
  `step_template._job_creation_metadata.extends_symtab` — the same callable
  `instantiate_model` uses. It depends only on the step's name, its `let` bindings
  and the job-scope table, so re-invoking it is deterministic and cannot drift from
  what the job was instantiated with.
- **Serialization goes through the engine.** `symtab_to_expr_values` for the typed
  coercion, then `SerializedSymbolTable.from_symtab` — the same serializer
  openjd-rs uses.
- **Imports stay lazy.** `openjd.expr` is imported inside the helper and under
  `TYPE_CHECKING`, so importing `openjd.model` still does not load the Rust
  bindings. Your `test_importing_parser_does_not_load_rust_expr_surface` caught my
  first attempt at this — thanks, it's a good test.
- **Not filtered.** openjd-rs narrows each step's table to the symbols that step
  references. This returns the full scope, which is a valid superset: a session
  layers its own scopes on top either way. Filtering would be a payload
  optimization and I did not want to reimplement `filter_symtab_for_step` here.

## Parity with openjd-rs

The point of this is that a v0 producer can feed a Rust consumer, so I measured
that rather than assuming it. 18 templates, 21 step comparisons, diffed entry by
entry on name, type and value against `openjd-rs`' `Step.resolved_symtab`:

**Zero missing symbols, zero semantic mismatches.** Cases covered: scalar types,
`BOOL`, `LIST[INT]`, `LIST[STRING]`, `LIST[PATH]`, `RANGE_EXPR`, PATH
(`RawParam`-only convention), template-scope `let` (frozen), script-scope `let`
(inputs only, results absent), `parameterSpace` ranges from a `let`,
`hostRequirements` from a `let`, `CHUNK[INT]`, step environments, multi-step
scoping, explicit parameter values, negative/zero/empty values, and the EXPR
function library.

`RANGE_EXPR` was the case I expected to break, since `symtab_to_expr_values`
omits it from `types` so the engine infers it from the value. Both sides produce
`('range_expr', '1-9:2')`, including the same `1-10:2` → `1-9:2` normalization.

`test_matches_the_rust_implementation` encodes this as a regression test.

**One cosmetic difference worth flagging.** For an integral-valued float written
`1.0` or `1.00`, openjd-rs emits `"1"` and this Python path emits `"1.0"`. Both
decode to the same `ExprValue`, so it is a transport-text difference rather than a
value one — confirmed by round-tripping each through `to_symtab` and comparing.

It still has a consequence. `SerializedSymbolTable` derives `Eq` and `Hash`
structurally over the JSON, so anything comparing two serialized tables byte-wise
sees a false mismatch on integral floats. Callers should compare decoded
`SymbolTable`s instead. Normalizing float display between the two implementations
would close it at the source, but that changes transport output, which felt out of
scope here.

## Testing

`hatch run test` — 5478 passed, 24 skipped, 3 xfailed. `hatch run lint` clean
(ruff, black, mypy). `cargo fmt --check` and `cargo check` clean.

Total coverage moves from 93.91% to 93.98%; every line added is covered. The 94%
gate was already short before this change.

17 new tests: 9 for the JSON accessors in
`test/openjd/expr/test_symbol_table.py`, 8 for the symbol tables in
`test/openjd/model_v0/test_create_job.py`.

## Two things to flag

**`scripts/generate_stubs.sh` does not run on current `mainline`** for me, so I
hand-wrote the two stub entries in `_openjd_rs.pyi` to match the generator's style
and verified with `mypy`. The script clones `pyo3-stub-gen` at `--depth 1` with no
pinned revision, and today's tip no longer compiles against this repo's pyo3:

```
error[E0599]: no method named `downcast` found ... help: there is a method `cast`
error: could not compile `pyo3-stub-gen` (lib) due to 8 previous errors
```

Pinning that clone to a known-good tag would make stub regeneration reproducible.
Happy to send it separately.

**`hatch run test` fails if `CONDA_PREFIX` is set** alongside `VIRTUAL_ENV`
(`maturin failed: Both VIRTUAL_ENV and CONDA_PREFIX are set`). Not caused by this
change, and easy to work around with `env -u CONDA_PREFIX`, but it bites anyone
developing from a conda base environment.

### DIFF
--- a/rust-bindings/src/expr/symbol_table.rs
+++ b/rust-bindings/src/expr/symbol_table.rs
@@ -269,6 +269,45 @@ impl PySerializedSymbolTable {
         }
     }
 
+    /// Build a ``SerializedSymbolTable`` from its JSON transport
+    /// text, as produced by ``to_json_str``.
+    ///
+    /// This is the inverse of ``to_json_str`` and exists for callers
+    /// that carry the transport form across a process or service
+    /// boundary — a scheduler persisting the table produced by
+    /// ``create_job`` and later handing it to a worker, for instance.
+    /// Prefer ``from_symtab`` when the source is an in-memory
+    /// ``SymbolTable``.
+    ///
+    /// Raises ``ValueError`` if the text is not valid JSON. Note that
+    /// the *contents* are validated lazily: a well-formed JSON
+    /// document whose entries are not valid symbol table entries is
+    /// accepted here and rejected by ``to_symtab``.
+    #[classmethod]
+    fn from_json_str(_cls: &Bound<'_, pyo3::types::PyType>, json: &str) -> PyResult<Self> {
+        let inner = openjd_expr::SerializedSymbolTable::from_json_str(json).map_err(|e| {
+            pyo3::exceptions::PyValueError::new_err(format!(
+                "Failed to parse SerializedSymbolTable JSON: {e}"
+            ))
+        })?;
+        Ok(Self { inner })
+    }
+
+    /// Serialize to the JSON transport text: an array of
+    /// ``{"name", "type", "value"}`` objects in canonical
+    /// (lexicographic) path order.
+    ///
+    /// Use this to move a table across a process or service boundary;
+    /// pair it with ``from_json_str`` to reconstruct. The result is
+    /// stable for a given table, so it is safe to store or compare.
+    fn to_json_str(&self) -> PyResult<String> {
+        serde_json::to_string(&self.inner).map_err(|e| {
+            pyo3::exceptions::PyValueError::new_err(format!(
+                "Failed to serialize SerializedSymbolTable: {e}"
+            ))
+        })
+    }
+
     /// Deserialize this serialized symbol table into a full
     /// ``SymbolTable`` suitable for inspection or modification.
     /// ``path_format`` controls how PATH-typed values are
--- a/src/openjd/_openjd_rs.pyi
+++ b/src/openjd/_openjd_rs.pyi
@@ -2032,6 +2032,36 @@ class SerializedSymbolTable:
         use this classmethod to convert.
         """
 
+    @classmethod
+    def from_json_str(cls, json: builtins.str) -> SerializedSymbolTable:
+        r"""
+        Build a ``SerializedSymbolTable`` from its JSON transport
+        text, as produced by ``to_json_str``.
+
+        This is the inverse of ``to_json_str`` and exists for callers
+        that carry the transport form across a process or service
+        boundary — a scheduler persisting the table produced by
+        ``create_job`` and later handing it to a worker, for instance.
+        Prefer ``from_symtab`` when the source is an in-memory
+        ``SymbolTable``.
+
+        Raises ``ValueError`` if the text is not valid JSON. Note that
+        the *contents* are validated lazily: a well-formed JSON
+        document whose entries are not valid symbol table entries is
+        accepted here and rejected by ``to_symtab``.
+        """
+
+    def to_json_str(self) -> builtins.str:
+        r"""
+        Serialize to the JSON transport text: an array of
+        ``{"name", "type", "value"}`` objects in canonical
+        (lexicographic) path order.
+
+        Use this to move a table across a process or service boundary;
+        pair it with ``from_json_str`` to reconstruct. The result is
+        stable for a given table, so it is safe to store or compare.
+        """
+
     def to_symtab(self, *, path_format: typing.Optional[PathFormat] = None) -> SymbolTable:
         r"""
         Deserialize this serialized symbol table into a full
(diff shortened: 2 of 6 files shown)


seant-aws (approved):
## PR #332 Review: `create_job_with_symbol_tables` + JSON accessors

**Verdict: Approve ✅ — good to merge.**

---

### Executive Summary

Clean, well-designed PR that adds two features we directly consume:
1. `SerializedSymbolTable.from_json_str()` / `.to_json_str()` — the public JSON transport API (replaces the private `_reconstruct_serialized_symtab` we're currently using in #1063)
2. `create_job_with_symbol_tables()` — DataPlane-side producer returning resolved tables per-step

Architecture is sound, 17 tests cover the critical paths, and the PR was verified against openjd-rs with zero semantic mismatches across 21 step comparisons.

---

### No Critical Issues

The reviewer bot flagged the `to_json_str` Rust method as potentially missing a return — **this is a false positive**. The actual code uses a tail expression (`serde_json::to_string(&self.inner).map_err(...)`) with no semicolons, which is valid Rust implicit return of `PyResult<String>`.

---

### Important Issues (should discuss, not blocking)

| # | Issue | Severity | Assessment |
|---|-------|----------|------------|
| 1 | **Deferred validation in `from_json_str`** — validates JSON syntax only, pushes content errors to `to_symtab()` call time | Important | This is fine for our consumption path because we call `to_symtab()` immediately after deserializing. But worth a docstring clarifying that callers should validate early. |
| 2 | **Bot comment #2 — stability claim scope** | Minor | The "canonical order" guarantee holds for `from_symtab`-built instances but not `from_json_str`-derived ones. A docstring scoping the claim would help. |

seant-aws:
Also have this as a followup: https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1063

mwiebe (approved):

PR #350, opened 2026-09-03
fix: Make task parameter type names case-insensitive under EXPR

leongdl:
## What changed

Template Schemas §2 says "job parameter and task parameter type names become case-insensitive" when the EXPR extension is enabled, and so are case-sensitive without it. `_normalize_parameter_type_case` already implemented that, gated on the extension, but it was registered only on `parameterDefinitions` in `JobTemplate` and `EnvironmentTemplate`. `StepParameterSpaceDefinition` had no such registration, so a task parameter spelled `type: int` was rejected even with EXPR declared:

```
steps[0] -> parameterSpace -> taskParameterDefinitions[0]:
	Input tag 'int' found using 'type' does not match any of the expected tags: ...
```

The fix is the same registration the other two models carry, on `taskParameterDefinitions`, before discriminated-union resolution. Nothing else was needed: `TaskParameterList` is a list of dicts, the shape the shared function was already written for, and `context.extensions` holds the effective set by then because `JobTemplate` declares `extensions` before `steps`. That covers all five task parameter types, `CHUNK[INT]` included — §3.4.1.5 makes it a task parameter type name, so EXPR makes `chunk[int]` a spelling of it.

The fold now uses an ASCII-only translation table instead of `str.upper()`. `str.upper()` is Unicode-aware and folds ten codepoints wholly into the type-name alphabet, five of which spell a real type: `ıNT` (U+0131), `ſTRING` (U+017F), `ﬂOAT` (U+FB02) and `ﬅRING`/`ﬆRING` (U+FB05/U+FB06) were all accepted under EXPR. Spec type names are ASCII, so none of those is a spelling variant. This is a second defect rather than part of the fixture's, and no fixture pins it, but it shares the line being edited — and registering the fold on task parameters would otherwise have extended it to a surface that was accidentally safe by not folding at all.

## Conformance

The fixture is `EXPR/job_templates/proposed/3.4.1--task-param-type-case-insensitive.yaml`, added by OpenJobDescription/openjd-specifications#166, which failed here and in openjd-rs alike. Both are `job_templates/` fixtures, so what they assert is template validity — what `openjd check` gets from `decode_job_template`. Loading the real fixture files and decoding them, before and after this commit:

| Fixture | Expected | Before | After |
|---|---|---|---|
| `EXPR/…/proposed/3.4.1--task-param-type-case-insensitive.yaml` | valid | invalid, `Input tag 'pAtH'` | valid |
| `base/…/proposed/2--type-lowercase.invalid.yaml` | invalid | invalid | invalid |
| `EXPR/job_templates/2--type-case-insensitive.yaml` (already gating) | valid | valid | valid |

The fixture this package was failing now passes, the base fixture it already satisfied is unchanged, and the already-green gating fixture stays green. The conformance runner itself was not executed, because that needs openjd-cli and openjd-sessions built against the patched model.

The fixture stays in `proposed/` until openjd-rs ships the matching change, since a fixture is promotable only when it passes on both implementations.

## Tests

81 tests across three classes, one per registration site, each covering the four cases of EXPR on/off by canonical/mis-cased spelling.

| Class | File | Surface |
|---|---|---|
| `TestTaskParameterTypeNameCase` | `test_parameter_space.py` | `StepParameterSpaceDefinition.taskParameterDefinitions` |
| `TestJobParameterTypeNameCase` | `test_list_parameters.py` | `JobTemplate.parameterDefinitions` |
| `TestEnvironmentTemplateParameterTypeNameCase` | `test_environment_template.py` | `EnvironmentTemplate.parameterDefinitions` |

Nine mutations, each caught by the tests that should catch it:

| Mutation | Caught by |
|---|---|
| un-register on `taskParameterDefinitions` | `test_with_expr_miscased_accepted` ×5 + 3 |
| drop `mode="before"` | the same 8 |
| un-register on `JobTemplate` | job `test_with_expr_miscased_accepted` ×4 |
| un-register on `EnvironmentTemplate` | env `test_with_expr_miscased_accepted` ×3 |
| drop the EXPR gate | `test_no_expr_miscased_rejected` ×12 + 2 |
| revert to `str.upper()` | `test_non_ascii_lookalike_rejected_with_expr` ×11 |
| no-op translation table | 15 |
| remove both `isinstance` guards | `test_malformed_input_is_a_validation_error_not_a_crash` ×6 |
| fold only the first list element | the two interaction tests |

Two groups survive every mutation and are negative controls rather than false pins. `test_no_expr_canonical_case_accepted` proves a mis-cased rejection is about the spelling and not about the type being unavailable. `test_non_ascii_lookalike_rejected_without_expr` is inert by construction, since without EXPR no fold runs; it exists so the pair covers both extension states, and its comment says so.

Full suite: 5626 passed, against a 5545 baseline. The 12 failures are pre-existing and unrelated — all in `test/openjd/model_v1/test_step_param_space_iter.py::TestChunksTaskCountOverride`, one cause, a stale Rust crate pin in the v1 binding layer. Byte-identical failure set before and after. `ruff`, `black` and `mypy` clean.

## What test auditing found that the design missed

Three findings came from auditing the tests rather than the code, and each added tests that were missing.

**The `EnvironmentTemplate` registration was covered by nothing.** Neutering it left the entire `model_v0` suite green at 2743 passed, where neutering the `JobTemplate` one fails 6 tests. The ASCII change had therefore silently altered that shipped path in both directions with no test either way.

**The shared function's two `isinstance` guards became load-bearing on a new surface.** Registering the fold routed `taskParameterDefinitions` through them for the first time. Measured with the guards removed: `taskParameterDefinitions:` left empty raises `TypeError: 'NoneType' object is not iterable`, `type: 3` raises `AttributeError: 'int' object has no attribute 'translate'`, and a missing `type` raises `KeyError`, each escaping `decode_job_template` instead of arriving as a `DecodeValidationError`.

**A mis-cased name now reaches validators it never used to.** Before the fold ran on this field, `type: int` failed at discriminator resolution and nothing downstream saw it. Now it resolves, so two mis-cased `chunk[int]` parameters report the one-`CHUNK[INT]`-per-step rule and `int`/`INT` under one name reports the duplicate-name rule. Both are correct and both are new, so both are asserted.

One test was renamed because mutation testing showed it did not pin what its name claimed. A template declaring EXPR to a caller that does not allowlist it is rejected on the extension name, before the type name is reached, so it survives removal of the EXPR gate. It now says so and asserts the extension message.

## Out of scope

Pydantic's wording for a type name that is wrong for reasons other than case (`Input tag 'NOPE' found using 'type' does not match any of the expected tags`) is unhelpful but is not this defect, and is not worth diverging from the framework to improve.

`openjd.model._v1` is entirely re-exports from the `openjd._openjd_rs` extension module, so that surface is fixed by the openjd-rs change rather than here.

## Related

- Design note: `SuperDaveDocs docs/conformance-0901/4.3-parameter-insensitive/fix.md`
- Conformance fixture: OpenJobDescription/openjd-specifications#166
- The matching openjd-rs change is in progress and will be linked here when raised.

### DIFF
--- a/src/openjd/model/v2023_09/_model.py
+++ b/src/openjd/model/v2023_09/_model.py
@@ -1829,6 +1829,13 @@ class StepParameterSpaceDefinition(OpenJDModel_v2023_09):
         reshape_field_to_dict={"taskParameterDefinitions": "name"},
     )
 
+    # §2 makes task parameter type names case-insensitive under EXPR, the same as
+    # job parameter type names. Must run before discriminated-union resolution.
+    @field_validator("taskParameterDefinitions", mode="before")
+    @classmethod
+    def _normalize_parameter_type_case(cls, v: Any, info: ValidationInfo) -> Any:
+        return _normalize_parameter_type_case(v, info)
+
     @field_validator("taskParameterDefinitions")
     @classmethod
     def _validate_parameters(cls, v: TaskParameterList) -> TaskParameterList:
@@ -3948,11 +3955,18 @@ def _expr_param_gate(value: JobParameterType, info: ValidationInfo) -> JobParame
     return value
 
 
+# Parameter type names are ASCII (§2). str.upper() is Unicode-aware and folds
+# U+0131 to 'I', which would make 'ıNT' a spelling of 'INT'.
+_ASCII_UPPERCASE = str.maketrans("abcdefghijklmnopqrstuvwxyz", "ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+
 def _normalize_parameter_type_case(value: Any, info: ValidationInfo) -> Any:
     """Uppercase the ``type`` discriminator of each parameter definition when
     the EXPR extension is enabled (RFC 0007 makes parameter type names
     case-insensitive, e.g. ``int`` == ``INT``, ``list[int]`` == ``LIST[INT]``).
     Runs before discriminated-union resolution.
+
+    Applies to job parameter and task parameter type names alike, per §2.
     """
     context = cast(Optional[ModelParsingContext], info.context)
     if not (context and "EXPR" in context.extensions):
@@ -3962,7 +3976,7 @@ def _normalize_parameter_type_case(value: Any, info: ValidationInfo) -> Any:
     normalized: list[Any] = []
     for item in value:
         if isinstance(item, dict) and isinstance(item.get("type"), str):
-            item = {**item, "type": item["type"].upper()}
+            item = {**item, "type": item["type"].translate(_ASCII_UPPERCASE)}
         normalized.append(item)
     return normalized
 
--- a/test/openjd/model_v0/v2023_09/test_environment_template.py
+++ b/test/openjd/model_v0/v2023_09/test_environment_template.py
@@ -5,6 +5,7 @@ from typing import Any
 import pytest
 from pydantic import ValidationError
 
+from openjd.model import DecodeValidationError, decode_environment_template
 from openjd.model._parse import _parse_model
 from openjd.model.v2023_09 import EnvironmentTemplate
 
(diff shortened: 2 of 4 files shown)


src/openjd/model/v2023_09/_model.py
```diff
@@ -1828,6 +1828,13 @@ class StepParameterSpaceDefinition(OpenJDModel_v2023_09):
         reshape_field_to_dict={"taskParameterDefinitions": "name"},
     )
 
+    # §2 makes task parameter type names case-insensitive under EXPR, the same as
+    # job parameter type names. Must run before discriminated-union resolution.
+    @field_validator("taskParameterDefinitions", mode="before")
+    @classmethod
+    def _normalize_parameter_type_case(cls, v: Any, info: ValidationInfo) -> Any:
+        return _normalize_parameter_type_case(v, info)
```
leongdl, line 1836: Just applies the validator below to support mixed case.

src/openjd/model/v2023_09/_model.py
```diff
@@ -3967,11 +3974,18 @@ def _expr_param_gate(value: JobParameterType, info: ValidationInfo) -> JobParame
...
 def _normalize_parameter_type_case(value: Any, info: ValidationInfo) -> Any:
     """Uppercase the ``type`` discriminator of each parameter definition when
     the EXPR extension is enabled (RFC 0007 makes parameter type names
     case-insensitive, e.g. ``int`` == ``INT``, ``list[int]`` == ``LIST[INT]``).
     Runs before discriminated-union resolution.
+
+    Applies to job parameter and task parameter type names alike, per §2.
     """
     context = cast(Optional[ModelParsingContext], info.context)
     if not (context and "EXPR" in context.extensions):
```
leongdl, line 3972: EXPR only - returns if `EXPR` is not in the extensions.

crowecawcaw (approved):

mwiebe (approved):

PR #352, opened 2026-09-03
fix: coerce LIST[BOOL] items per RFC 0007 §2.15

wyongzhi:
## What was the problem/requirement? (What/Why)

Conformance fixtures `2.15--list-bool-param-{interpolation,runtime}` (added by OpenJobDescription/openjd-specifications#158) pass on openjd-rs but fail on the Python CLI with `List contains incompatible types: bool, string`.

Template Schemas §2.15 states each LIST[BOOL] item accepts the same values as scalar BOOL: bool literals, int/float `0`/`1`, and the case-insensitive strings `true`/`yes`/`on`/`1` and `false`/`no`/`off`/`0`. The Python model validated those spellings at decode time but stored the list verbatim, so a heterogeneous list reached the expression engine un-normalized. Homogeneous non-bool spellings (e.g. `["yes","off"]`, `[1,0]`) were worse: silently accepted as the wrong list type (`ListString`/`ListInt`), failing only when a boolean operator touched an element.

## What was the solution? (How)

Coerce each LIST[BOOL] item to a canonical bool at the job-creation boundary (`_create_job.py` only; template models are not mutated):

```
YAML template
   |  decode (pydantic validates spellings, stores them verbatim; unchanged)
   v
Template object    default = ["yes", 0, True]    <- keeps the author's spellings (pinned by test)
   |
   |  create_job()  * conversion happens at this gate
   v
Job object         value = [True, False, True]   <- every reader sees canonical bools
   |
   v
Expression engine / sessions / cli               <- untouched, fixed for free
```

- template defaults: `_collect_defaults_2023_09`
- submitted values: `_coerce_expr_param_value`, both native lists and JSON strings (the shared LIST JSON parse is unchanged for other list types)
- gated exactly on `LIST_BOOL`; invalid items keep `Parameter <name>:` context

The coercion reuses `_coerce_bool_value`, the same function the scalar BOOL type and the decode-time item validation use, so the accepted value set cannot drift between the scalar and list types.

**Why convert at job creation, not at template decode (where openjd-rs converts):** the difference is where the conversion can live. In Rust it lives in the type system: `BoolValue` deserializes `"yes"` straight into a bool, so an un-normalized value is unrepresentable. In pydantic it would live in validators, and validators can be bypassed: environment-template merges go through `model_copy`, which skips them, so decode-time normalization leaks that path and the job-creation coercion is needed anyway. One site instead of two. It also keeps decoded templates round-tripping the author's spellings, a long-released behavior. Both implementations converge where conformance observes them: the created Job's values.

LIST[FLOAT] integer items are intentionally not normalized: §2.14 enumerates no alternate spellings, no fixture pins it, and the expression engine promotes int to float. Noting it here rather than changing it silently.

## What was the impact?

Created-Job LIST[BOOL] values are now canonical booleans. A previously-working homogeneous `["yes","no"]` now stores and interpolates as `true`/`false`, matching openjd-rs. Audited openjd-sessions, openjd-cli, and downstream consumers: none read the raw spellings.

Template objects are unchanged: a decoded template's `default` keeps the author's original spellings.

## Testing

- 31 new unit tests: mixed/homogeneous/mixed-case spellings, native and JSON submitted values, adversarial items via both the JSON and native-list paths (`[[true]]`, `[null]`, `[2]`, `[2.0]`, non-list JSON), empty list, environment-template merged defaults, `Parameter <name>:` context on per-item failures (JSON-level errors stay verbatim like other list types), template-object non-mutation and dump round-trip, `default=None` guard, and non-BOOL list guards (LIST[INT] JSON parsing, LIST[STRING]/LIST[PATH]/LIST[LIST[INT]] passthrough). Full suite: 5648 passed; ruff/black/mypy clean.
- Conformance A/B (`run_openjd_cli_tests.py`, full 2023-09 suite, released 0.11.6 vs this change): 1172 passed / 2 failed → 1174 passed / 0 failed. The only delta is the two 2.15 list-bool fixtures flipping to pass:

```
240,241c240,241
<   ✗ 2.15--list-bool-param-interpolation
<   ✗ 2.15--list-bool-param-runtime
---
>   ✓ 2.15--list-bool-param-interpolation
>   ✓ 2.15--list-bool-param-runtime
```

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license.*

### DIFF
--- a/src/openjd/model/_bool_coercion.py
+++ b/src/openjd/model/_bool_coercion.py
@@ -1,0 +1,36 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from typing import Any
+
+# Accepted string spellings for boolean defaults/values (case-insensitive),
+# per RFC 0007 (BOOL parameter type).
+_BOOL_TRUE_STRINGS = frozenset({"true", "yes", "on", "1"})
+_BOOL_FALSE_STRINGS = frozenset({"false", "no", "off", "0"})
+
+
+def _coerce_bool_value(value: Any) -> bool:
+    """Coerce an RFC 0007 BOOL value to a Python bool, raising ValueError for
+    anything outside the accepted set (bool, int 0/1, float 0.0/1.0, or a
+    case-insensitive true/false/yes/no/on/off/1/0 string).
+    """
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, int):  # bool already handled above
+        if value in (0, 1):
+            return bool(value)
+        raise ValueError("BOOL value as an integer must be 0 or 1.")
+    if isinstance(value, float):
+        if value in (0.0, 1.0):
+            return bool(value)
+        raise ValueError("BOOL value as a float must be 0.0 or 1.0.")
+    if isinstance(value, str):
+        low = value.lower()
+        if low in _BOOL_TRUE_STRINGS:
+            return True
+        if low in _BOOL_FALSE_STRINGS:
+            return False
+        raise ValueError(
+            "BOOL value as a string must be one of (case-insensitive): "
+            "true, false, yes, no, on, off, 1, 0."
+        )
+    raise ValueError("BOOL value must be a boolean, 0/1, 0.0/1.0, or a boolean string.")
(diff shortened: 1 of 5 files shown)


jericht (approved):

seant-aws (approved):

seant-aws:
looks good !

PR #367, opened 2026-09-16
fix: Allow underscores in a combination expression

leongdl:
Fixes: *n/a — found while auditing `CombinationExpression` validation coverage in a downstream consumer*

Template Schemas §3.4.3 constraint 1 gives a `<CombinationExpr>` "Any [character] allowed in an `<Identifier>` plus the space, `*`, `(`, and `)` characters", and §7.1 lists the underscore explicitly among an `<Identifier>`'s characters. The character class in `v2023_09/_model.py` omitted it:

```python
pattern=r"(?-m:^[A-Za-z0-9\*\(\), ]+\z)"
```

So a task parameter named `Frame_Range` could be declared, but never referenced from a combination expression:

```
steps[0] -> parameterSpace -> combination:
	String should match pattern '(?-m:^[A-Za-z0-9\*\(\), ]+\z)'
```

The rejection landed on the pattern, before `_internal/_combination_expr.Parser` ran — that parser accepted underscores all along, since its `IdentToken` is `^[^\d\W][\w]*$`. The comment above the constraint already claimed the intended relationship to `<Identifier>`; only the class disagreed.

The two decode paths therefore disagreed. Measured on `Frame_Range * Quality` with both parameters declared:

| Path | Before | After |
|---|---|---|
| `parse_model(model=JobTemplate_2023_09, ...)` | rejected on the pattern | accepted |
| `openjd._openjd_rs.decode_job_template_str` | accepted | accepted |

`openjd-rs` never had the defect: `validate_v2023_09/structure.rs` has `_` in its allow-list and its tokenizer accepts it in names, so `openjd check` passes the same template and reports the full 6-task space.

Bind the identifier character class to one constant and build both patterns from it, rather than adding `_` to the literal and leaving the duplication that let the two drift:

```python
_identifier_chars = r"A-Za-z0-9_"
_identifier_regex = rf"(?-m:^[A-Za-z_][{_identifier_chars}]*\z)"
...
pattern=rf"(?-m:^[{_identifier_chars}\*\(\), ]+\z)",
```

The rendered `_identifier_regex` is byte-identical to before, checked at runtime, so `Identifier` behaviour is unchanged.

A combination expression may now reference a task parameter whose name contains an underscore, which is every name §7.1 permits. Templates previously rejected are now accepted; nothing previously accepted is now rejected. Consumers that validate a template through this library and had worked around the restriction by renaming parameters no longer need to.

- Have you run the unit tests? **Yes.** `hatch run test`: 6069 passed, 24 skipped, 3 xfailed, coverage gate met. `hatch run lint` (ruff, black, mypy) clean.

New tests, all mutation-checked:

- `TestCombinationExprCharacterClass` in `test/openjd/model_v0/v2023_09/test_parameter_space.py` — six accepted cases (interior, leading and trailing underscore, a bare `_`, and one inside an association), four negative controls (`-`, `.`, `+`, and a tab, which is not the permitted space) asserting the full pattern message, and one end-to-end case decoding a template, creating the job, and asserting `StepParameterSpaceIterator` yields the 3 × 2 space keyed by `Frame_Range`.
- `test_combination_accepts_underscore_names` in `test/openjd/model_v1/test_step_param_space_def.py` — control pinning the Rust path's half of the parity, so a future tightening there cannot silently reopen the divergence.

Two mutants, each applied to the source and the suite re-run:

| Mutant | Result |
|---|---|
| Revert the class to the hardcoded `[A-Za-z0-9\*\(\), ]` | caught — all 10 fail |
| Widen the class to `.+` | caught — exactly the 4 negative controls fail, the 6 positives stay green |

- Are relevant docstrings in the code base updated? **Yes** — the comment above `CombinationExpr` now says why the class is shared rather than restating the relationship it previously only asserted.

There is no conformance fixture covering this today. `2023-09/base/job_templates/` carries a §3.4.3 family including `3.4.3--disallowed-char-at-sign.invalid.yaml`, but nothing positive on the underscore, which is why the divergence went unnoticed. Both CLIs now agree, so a fixture is warranted; I will raise it separately against `openjd-specifications`.

Separately, §3.4.3 constraint 1 omits the comma from its operator list even though `<ExprList>` in the grammar directly above requires one. Both implementations allow it. That reads as a spec erratum and I will raise it too.

No. The accepted input set only grows.

No. The change widens a character class to the set the specification defines, and the widened class remains a strict subset of the identifier and operator characters — the negative controls pin that a hyphen, dot, plus and tab are all still rejected. No file or directory handling is involved.

----


src/openjd/model/v2023_09/_model.py
```diff
@@ -1778,11 +1780,16 @@ def _validate_range_elements(cls, value: Any) -> Any:
...
+# the two cannot drift: a task parameter name this rejects could never be
+# referenced, which is what happened while '_' was missing here.
 CombinationExpr = Annotated[
     str,
     StringConstraints(
-        min_length=1, max_length=1280, strict=True, pattern=r"(?-m:^[A-Za-z0-9\*\(\), ]+\z)"
+        min_length=1,
+        max_length=1280,
+        strict=True,
+        pattern=rf"(?-m:^[{_identifier_chars}\*\(\), ]+\z)",
```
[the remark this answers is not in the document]
leongdl: Splitting this: the whitespace facts are correct, the prescription is not, and the point about the comment is fair. Addressed in b402d57.

**The divergence is real, and the pattern is the side to keep.** `_tokenstream.py:107-110` does fold all whitespace, and a `|` literal block does reach the pattern with its newline intact. But the missing oracle is what the other implementation does. Measured on `Frame_Range *\nQuality`:

| | newline | tab |
|---|---|---|
| v0 pydantic pattern | rejected | rejected |
| v1, `decode_job_template_str` (openjd-rs) | rejected, `contains disallowed character` | rejected, same |
| bare `CombinationExpressionParser` | accepted | accepted |

So this is not a cross-implementation divergence, it is internal to this package, and the pattern already agrees with openjd-rs. §3.4.3 constraint 1 says "plus the space" and never mentions a multi-line expression, so widening the class to `\s` would *create* a divergence rather than close one. Taking your first resolution: the pattern is correct, the tokenizer is over-permissive for this grammar.

Worth adding that `TokenStream.__init__` is shared with the format-string path, where multi-line genuinely is allowed. That comment is about the tokenizer in general, not a claim about combination expressions, which is why it reads as contradicting the pattern.

One correction: a folded `>` scalar does not survive either. YAML's default clip chomping keeps a trailing newline, so `>` yields `'Frame_Range * Quality\n'` and that is rejected by the pattern as well. Only `>-` produces `'Frame_Range * Quality'` and passes. So there is no accidental escape hatch, which makes leaving the pattern narrow cheaper than it looked.

**On the comment.** You are right that "the two cannot drift" over-promises. Only the character class is shared, not the leading-character rule, and a digit-leading token does get past the pattern and fall over in the lexer. Rewritten to say what is actually shared and to record the U+0020 decision:

```python
# §3.4.3: an Identifier's characters plus the space and the operators. Only the
# character class is shared with Identifier, not its leading-character rule.
# The space is deliberately just U+0020. The shared TokenStream folds all
# whitespace, which is wider than §3.4.3 allows, so a newline is refused here.
```

And rather than leave the whitespace decision implicit, `test_disallowed_characters_still_rejected` now carries a `newline` case alongside the `tab` one, with the reasoning in the class docstring. A mutant widening the space to `\s` is caught by exactly those two cases.

test/openjd/model_v0/v2023_09/test_parameter_space.py
```diff
@@ -1071,3 +1077,114 @@ def test_two_miscased_chunk_int_report_the_one_chunk_rule(self) -> None:
...
+        with pytest.raises(ValidationError) as excinfo:
+            _parse_model(
+                model=StepParameterSpaceDefinition,
+                obj=self._space(["Frame", "Range", "Quality"], combination),
+            )
+
+        # THEN
+        message = str(excinfo.value)
+        assert "combination" in message, message
+        assert r"String should match pattern '(?-m:^[A-Za-z0-9_\*\(\), ]+\z)'" in message, message
```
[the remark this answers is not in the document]
leongdl: Correct, and it was already visible in my own mutation run: reverting the constant failed all ten tests, including the four negative controls, which have nothing to do with the underscore. That coupling is the smell you are describing. Fixed in b402d57, taking your suggestion close to verbatim:

```python
assert "combination" in message, message
assert "String should match pattern" in message, message
assert _identifier_chars in message, message
```

Verified rather than assumed, by adding a control case to the harness that widens `_identifier_chars` to `A-Za-z0-9_%` and requires the suite to stay green:

| Change to the class | Expected | Result |
|---|---|---|
| Revert to hardcoded, no `_` | caught | caught, 11 fail |
| Widen to `.+` | caught | caught, the 5 controls fail |
| Widen the space to `\s` | caught | caught, `tab` and `newline` fail |
| Legitimately widen the constant with `%` | stays green | 12 passed |

That last row is the property the old assertion did not have. Full suite 6070 passed, 24 skipped, 3 xfailed; ruff, black and mypy clean.

src/openjd/model/v2023_09/_model.py
```diff
@@ -1777,12 +1779,17 @@ def _validate_range_elements(cls, value: Any) -> Any:
...
+# The space is deliberately just U+0020. The shared TokenStream folds all
+# whitespace, which is wider than §3.4.3 allows, so a newline is refused here.
 CombinationExpr = Annotated[
     str,
     StringConstraints(
-        min_length=1, max_length=1280, strict=True, pattern=r"(?-m:^[A-Za-z0-9\*\(\), ]+\z)"
+        min_length=1,
+        max_length=1280,
+        strict=True,
+        pattern=rf"(?-m:^[{_identifier_chars}\*\(\), ]+\z)",
```
mwiebe, line 1792: Should this start with `[A-Za-z_]` before the identifier characters part?

seant-aws (approved):

src/openjd/model/v2023_09/_model.py
mwiebe: Or I see, this is more of a sanity check than verifying the grammar of combination expressions.

leongdl: Your follow-up is right, and it holds up when measured. The pattern is a character-class filter; the leading-character rule is enforced by the lexer downstream, so nothing digit-leading gets through:

| `combination` | v0 | v1 (openjd-rs) |
|---|---|---|
| `1A * B` | `Unexpected '1' in '1A * B' after ''` | `parameter 'A' missing from combination expression.` |
| `1 * A * B` | `Unexpected '1' in '1 * A * B' after ''` | `references unknown parameter '1'.` |
| `1` | `Unexpected '1' in '1' after ''` | `parameter 'A' missing from combination expression.` |
| `A1 * B` | accepted | accepted |

Worth noting the v0 diagnostic is the better one here: it quotes the offending character and the expression, where a pattern rejection would emit the raw regex. So anchoring the class would arguably make the message worse for this case without changing the outcome.

The automated reviewer raised the same point on this thread, and b402d57 tightened the comment so it no longer implies the leading-character rule is shared with `Identifier`:

```python
# §3.4.3: an Identifier's characters plus the space and the operators. Only the
# character class is shared with Identifier, not its leading-character rule.
```

Resolving on your retraction. Happy to add these four cases as a test if you would like the boundary pinned rather than just measured.

leongdl: One thing worth adding on scope, since the refactor makes the class more prominent than it was: **this PR does not change the leading-character behaviour at all.** The old literal had no anchor either, so `1A * B` got past the pattern before this change exactly as it does after.

Measured by running both patterns through pydantic directly:

| input | old pattern | new pattern | changed? |
|---|---|---|---|
| `1A * B` | accepts | accepts | no |
| `1 * A * B` | accepts | accepts | no |
| `A1 * B` | accepts | accepts | no |
| `Frame_Range * Quality` | rejects | accepts | **yes** |

The underscore row is the only behavioural change in the diff. Everything else is an extraction: `_identifier_regex` renders byte-identical to before, checked at runtime, so `Identifier` itself is untouched.

So the shape you spotted is pre-existing rather than introduced here. Naming the class is what made it visible, which seems like a fair trade.

mwiebe (approved):

